PR #38, opened 2012-10-03
The htmlform edit form textarea should resize based on the window size - Html 85

dkayiwa:
The htmlform edit form textarea should resize based on the window size - Html 85


dkayiwa:
Closing this because ticket was closed.

PR #79, opened 2015-02-12
Try this html574

mogoodrich:


api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
```diff
@@ -84,10 +89,36 @@ public void addAction(FormSubmissionControllerAction action) {
...
      * @param submission
      */
-    public void handleFormSubmission(FormEntrySession session, HttpServletRequest submission) {
+    public void handleFormSubmission(FormEntrySession session, HttpServletRequest submission) throws Exception{
         lastSubmission = submission;
-        for (FormSubmissionControllerAction element : actions) {
-            element.handleSubmission(session, submission);
+        //Serialize when opted in.
+        String optedIn = Context.getAdministrationService().getGlobalProperty("htmlformentry.archiveHtmlForms","No");
+        if(Boolean.parseBoolean(optedIn)) {
```
mogoodrich: I'd pull this serialization code out into another method--a private method that is still within this class is fine.  Just so for future developers it's clear that they can ignore this block of code if they aren't concerned with serialization.  As it is now, it kind of buries the meat of the method, which is the for (FormSubmissionControllerAction : actions) block.

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
```diff
@@ -84,10 +89,36 @@ public void addAction(FormSubmissionControllerAction action) {
...
+            //Try to serialize
+            try {
```
mogoodrich: actually, ignore my previous comment, I changed mine mind, this seems fine as-is.

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
```diff
@@ -118,5 +149,38 @@ public HttpServletRequest getLastSubmission() {
...
+     * submission
+     * @param submittedData  SerializableFormObject
+     * @throws Exception
+     */
+    protected void serializeFormData(final SerializableFormObject submittedData) throws Exception {
+        //Get archive Directory
+        String path = HtmlFormEntryUtil.getArchiveDirPath();
+
+        //Ignore if no path specified
+        if(path==null)return;
```
mogoodrich: Log an error here in this case?

api/src/main/java/org/openmrs/module/htmlformentry/SerializableFormObject.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * This method returns a FormEntrySession object using data in the instance object
+     * @return FormEntrySession object
+     * @throws Exception
+     */
+    private FormEntrySession createSession() throws Exception {
+        //TODO: Check for null patientUuid and try to parse the xml to obtain the patient ID instead.
+        Patient patient = Context.getPatientService().getPatientByUuid(getPatientUuid());
```
mogoodrich: Shouldn't we always be able to have a patient ID here?

mogoodrich:
This looks good @mhawila, though it will be to start adding some tests to see if it actually work... :)

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
mhawila: I didn't think of logging as serialization is optional. But now that I think of it I think I can log a DEBUG level message.

api/src/main/java/org/openmrs/module/htmlformentry/SerializableFormObject.java
mhawila: Yeah we could but I thought using uuid is a better practice.

dkayiwa:
@mhawila did you get a chance to add some tests? :)

mogoodrich:
@dkayiwa fwiw, there is also another outstanding issue around this documented on the ticket, 
HTML-574

dkayiwa:
@mhawila do you still have some time to finish this?

mhawila:
Sure I will work on it. I will be traveling the coming week. I should be able to fix the issue by Friday though.

Cheers.

dkayiwa:
It has been a while since this was opened. A lot has happened and i can see merge conflicts. So am closing it for now. I can always reopen if you want me to. :)

PR #80, opened 2015-03-12
HTML-574: Added global properties htmlformentry.archiveDir and htmlforme...

mhawila:
...ntry.archiveHtmlForms used to specify archive

directory and opt to archive the submitted forms respectively. Used xstream to serialize the submitted html form to xml file.

HTML-574: Added htmlformId and form to serializable properties

HTML-574: Implemented the API for reprocessing archived HTML forms

### DIFF
--- a/api-tests/src/test/java/org/openmrs/module/htmlformentry/FormSubmissionControllerTest.java
+++ b/api-tests/src/test/java/org/openmrs/module/htmlformentry/FormSubmissionControllerTest.java
@@ -1,0 +1,18 @@
+package org.openmrs.module.htmlformentry;
+
+/**
+ * The contents of this file are subject to the OpenMRS Public License
+ * Version 1.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * http://license.openmrs.org
+ * <p/>
+ * Software distributed under the License is distributed on an "AS IS"
+ * basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
+ * License for the specific language governing rights and limitations
+ * under the License.
+ * <p/>
+ * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
+ */
+
+public class FormSubmissionControllerTest {
+}
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
@@ -1,13 +1,20 @@
 package org.openmrs.module.htmlformentry;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.Encounter;
+import org.openmrs.Patient;
+import org.openmrs.api.APIException;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.htmlformentry.action.FormSubmissionControllerAction;
+import org.openmrs.module.htmlformentry.action.RepeatControllerAction;
+import org.openmrs.util.OpenmrsUtil;
+
+import javax.servlet.http.HttpServletRequest;
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-
-import javax.servlet.http.HttpServletRequest;
-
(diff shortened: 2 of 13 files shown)


api-tests/src/test/java/org/openmrs/module/htmlformentry/FormSubmissionControllerTest.java
```diff
@@ -0,0 +1,18 @@
...
+ * Software distributed under the License is distributed on an "AS IS"
+ * basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
+ * License for the specific language governing rights and limitations
+ * under the License.
+ * <p/>
+ * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
+ */
+
+public class FormSubmissionControllerTest {
+}
```
mogoodrich: Is this coming soon? :)

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
```diff
@@ -84,10 +89,37 @@ public void addAction(FormSubmissionControllerAction action) {
...
+            try {
+                Patient patient = session.getPatient();
+                Encounter encounter = session.getEncounter();
+                SerializableFormObject formObject;
+                if (patient != null && encounter != null) {
+                    formObject = new SerializableFormObject(session.getXmlDefinition(),
+                            submission.getParameterMap(), patient.getPatientIdentifier().getIdentifier(),
+                            patient.getUuid(), encounter.getUuid(),session.getHtmlFormId());
+                } else {
+                    formObject = new SerializableFormObject(session.getXmlDefinition(), submission.getParameterMap(),
```
mogoodrich: Sorry, I think I asked this before, but I forget... :)  what is the use case where actually don't have a patient and encounter?

api-tests/src/test/java/org/openmrs/module/htmlformentry/FormSubmissionControllerTest.java
mhawila: I think I was about to add some tests and then I realized there was nothing to add. I forgot to remove the class.

mogoodrich:
@mhawila this looks good, thanks! A couple questions in-line above, but that's it.

One thing that would be good to add is serialization test to go along with the deserialization test.

However, I'm more or less ready to merge this in--I'm just going to wait until next week/Monday if that's okay with you.  I never like doing a big merge late on a Friday and then walking away for the weekend... ;)

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionController.java
mhawila: I haven't seen any implementation that don't have patient and encounter.  In other words that is pretty much what there is to it. I might be missing something though and if it is the case please let me know.

mogoodrich: Right, so shouldn't it just fail in this case?

mhawila: I guess you are right. I remember you suggested throwing an exception, right?

mogoodrich: Probably just logging an error is fine

mhawila: I actually went back to the code and realized the reason I didn't test for this use case is the fact that at this point the form is validated and can be saved. I put the explicit check for null patient and encounter to avoid getting NPE just incase i.e. I was just being careful. In reality the only possible scenario at this point is at least a non null patient. I will add the use case anyway just for clarity.

mhawila:
@mogoodrich I am trying to see how I can mock a test for serialization but it comes as a bit challenging. Do you have some ideas? Thanks.

mogoodrich:
Sorry, nothing off the top of my head @mhawila re: serialization.  I spent today struggling with my dev environment, but hopefully I can merge this in tomorrow... I will out of the office the rest of the week.

mogoodrich:
I went ahead and merged!

PR #82, opened 2015-04-01
HTML-583: Updating jquery core and ui libraries to enable drugAnswer att...

mhawila:
...ribute of obs tag to work correctly


mogoodrich:
Thanks @mhawila!

PR #84, opened 2015-04-21
HTML-381: Added validation while processing obs and obsgroup

geoff-wasilwa:
HFE did not provide checks for obs used as set members and obsgroup (if the obsgroup concept was a set).
Added a mechanism to validate this and store concept ids violating this rule to FormEntryContext from where
they can be displayed on the HTML Form edit page.

### DIFF
--- a/api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsGroupTagHandlerTest.java
+++ b/api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsGroupTagHandlerTest.java
@@ -1,0 +1,55 @@
+package org.openmrs.module.htmlformentry.handler;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openmrs.module.htmlformentry.HtmlFormEntryUtil;
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+public class ObsGroupTagHandlerTest extends BaseModuleContextSensitiveTest {
+	
+	@Test
+	public void validate_shouldWarnWhenObsgroupConceptIsNotASet() throws Exception {
+		String xml = "<htmlform>\n"
+				+ "<obsgroup groupingConceptId=\"10\">\n"
+				+ "<obs conceptId=\"3\" />\n"
+				+ "<obs conceptId=\"4\" />\n"
+				+ "</obsgroup>\n"
+				+ "</htmlform>";
+		
+		Document document = HtmlFormEntryUtil.stringToDocument(xml);
+		Node obsGroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+		TagAnalysis analysis = new ObsGroupTagHandler().validate(obsGroupNode);
+		Assert.assertEquals(1, analysis.getWarnings().size());
+	}
+	
+	@Test
+	public void validate_shouldRejectXmlWhenObsgroupGroupingConceptIdIsMissing() throws Exception {
+		String xml = "<htmlform><obsgroup>TEST</obsgroup></htmlform>";
+		TagAnalysis analysis = validateObsGroupTag(xml);
+		Assert.assertEquals(1, analysis.getErrors().size());
+	}
+	
+	@Test
+	public void validate_shouldRejectXmlWhenObsgroupConceptCannotBeFound() throws Exception {
+		String xml = "<htmlform><obsgroup groupingConceptId=\"none\">TEST</obsgroup></htmlform>";
+		TagAnalysis analysis = validateObsGroupTag(xml);
+		Assert.assertEquals(1, analysis.getErrors().size());
+	}
+	
+	@Test
+	public void validate_shouldPassWhenObsGroupConceptIsASet() throws Exception {
+		String xml = "<htmlfom><obsgroup groupingConceptId=\"23\"><obs conceptId=\"xx\" /></obsgroup></htmlfom>";
+		TagAnalysis analysis = validateObsGroupTag(xml);
+		Assert.assertEquals(0, analysis.getErrors().size());
+	}
+	
+	private TagAnalysis validateObsGroupTag(String xml) throws Exception {
+		Document document = HtmlFormEntryUtil.stringToDocument(xml);
+		Node obsGroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+		TagAnalysis analysis = new ObsGroupTagHandler().validate(obsGroupNode);
+		return analysis;
+	}
+	
+}
--- a/api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsTagHandlerTest.java
+++ b/api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsTagHandlerTest.java
@@ -1,0 +1,90 @@
+package org.openmrs.module.htmlformentry.handler;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openmrs.module.htmlformentry.HtmlFormEntryUtil;
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+public class ObsTagHandlerTest extends BaseModuleContextSensitiveTest {
+	
+	@Test
+	public void validate_shouldWarnWhenObsIsNotAMemberOfParentObsgroup() throws Exception {
+		String xml = "<htmlform>\n"
+				+ "<obsgroup groupingConceptId=\"23\">\n"
+				+ "<obs conceptId=\"4\" />\n"
+				+ "</obsgroup>\n </htmlform>";
+		Document document = HtmlFormEntryUtil.stringToDocument(xml);
+		Node parentObsGroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+		Assert.assertEquals("23", HtmlFormEntryUtil.getNodeAttribute(parentObsGroupNode, "groupingConceptId", null));
+		Node obsNode = HtmlFormEntryUtil.findChild(parentObsGroupNode, "obs");
+		TagAnalysis analysis = new ObsTagHandler().validate(obsNode);
+		Assert.assertEquals(1, analysis.getWarnings().size());	
+	}
+	
+	@Test
+	public void validate_shouldPassWhenObsIsAMemberOfParentObsgroup() throws Exception {
+		String xml = "<htmlform>\n"
+				+ "<obsgroup groupingConceptId=\"23\">\n"
+				+ "<obs conceptId=\"18\" />\n"
+				+ "	</obsgroup>\n </htmlform>";
+		
+		Document document = HtmlFormEntryUtil.stringToDocument(xml);
+		Node parentObsGroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+		Assert.assertEquals("23", HtmlFormEntryUtil.getNodeAttribute(parentObsGroupNode, "groupingConceptId", null));
+		
+		Node obsNode = HtmlFormEntryUtil.findChild(parentObsGroupNode, "obs");
+		TagAnalysis analysis = new ObsTagHandler().validate(obsNode);
+		Assert.assertEquals(0, analysis.getWarnings().size());
+	}
+	
+	@Test
+	public void validate_shouldRejectXmlWhenEitherObsConceptIdOrConceptIdsIsMissing() throws Exception {
+		String xml = "<htmlform><obs>TEST</obs></htmlform>";
+		TagAnalysis analysis = validateObsTag(xml);
+		Assert.assertEquals(1, analysis.getErrors().size());
+	}
+	
+	@Test
+	public void validate_shouldRejectXmlWhenBothObsConceptIdAndConceptIdsIsMissing() throws Exception {
+		String xml = "<htmlform><obs conceptId=\"xx\" conceptIds=\"xx,yy\">TEST</obs></htmlform>";
+		TagAnalysis analysis = validateObsTag(xml);
+		Assert.assertEquals(1, analysis.getErrors().size());
+	}
+	
+	@Test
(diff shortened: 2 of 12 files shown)


api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsGroupTagHandlerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void before() throws Exception {
+    	
+    	executeDataSet(XML_OPENMRS_TEST_DATASET);
+    	
+    	Patient demo = HtmlFormEntryUtil.getFakePerson();
+        formEntrySession = new FormEntrySession(demo, XML_DEFINITION, null);
+
+        out = new CapturingPrintWriter();
+        handler = new ObsGroupTagHandler();
+
```
mogoodrich: Can you clean up the tabs/formatting in this file? Thanks!

api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsGroupTagHandlerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void testSettingNonSetConceptAsObsGroupConcept() throws Exception {
+    	
+    	Document document = HtmlFormEntryUtil.stringToDocument(XML_DEFINITION);
+        Node obsgroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+        assertTrue(handler.doStartTag(formEntrySession, out, obsgroupNode.getParentNode(), obsgroupNode));
+        assertEquals(1, formEntrySession.getContext().getInvalidObsGroupConceptIds().size());
+    	
+    }
+	
+}
```
mogoodrich: Can you add a test to make sure that we *don't* get an error if the obs group is a set?

api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsTagHandlerTest.java
```diff
@@ -0,0 +1,50 @@
...
+	
+	@Test
+	public void testAddingNonSetMembersToObsGroup() throws Exception {
+		
+		Document document = HtmlFormEntryUtil.stringToDocument(XML_DEFINITION);
+		Node obsGroupNode = HtmlFormEntryUtil.findDescendant(document, "obsgroup");
+        Node obsNode = HtmlFormEntryUtil.findDescendant(document, "obs");
+        Boolean digDeeper = obsGroupHandler.doStartTag(formEntrySession, out, obsGroupNode.getParentNode(), obsGroupNode);
+        Boolean digDeeper2 = obsHandler.doStartTag(formEntrySession, out, obsNode.getParentNode(), obsNode);
+        assertEquals(1, formEntrySession.getContext().getInvalidSetMemberConceptIds().size());
```
mogoodrich: Nitpick, but if we aren't using the "dipDeeper" variables, can we just remove them?  The meaning of "digDeeper" isn't obviously apparent.

mogoodrich: Also, like in the above test, can we add an confirmation that we getInvalidSetMemberConceptIds() is null/0 if the form is properly defined?

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
```diff
@@ -96,6 +98,10 @@
     // TODO 1) change the type of this variable to visit
     // TODO 2) change HtmlFormEntryController so that it correctly populates the context with the relevent visit (if available)
     private Object visit;
+    
+    private List<Integer> invalidObsGroupConceptIds = new ArrayList<Integer>();
+    
+    private Map<Integer,List<Integer>> invalidSetMemberConceptIds = new HashMap<Integer,List<Integer>>();
```
mogoodrich: It seems kind of odd to make this error counters a property of the Form Entry Context, though the "right" location doesn't spring to mind--did you think about other places to handle this?

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -44,6 +47,18 @@ public boolean doStartTag(FormEntrySession session, PrintWriter out, Node parent
...
+		ObsGroup currentObsGroup = context.getActiveObsGroup();
+		Map<String, String> attributes = getAttributes(node);
+		if (currentObsGroup != null && currentObsGroup.getConcept().isSet() && attributes.containsKey("conceptId")) {
+			Concept obsConcept = HtmlFormEntryUtil.getConcept(attributes.get("conceptId"));
+			if (!currentObsGroup.getConcept().getSetMembers().contains(obsConcept)) {
+				context.addInvalidSetMemberConceptIds(currentObsGroup.getConcept().getConceptId(), obsConcept.getConceptId());
+			}
+		}
+	}
+
```
mogoodrich: I'd have preferred that this (and the validation test in ObsGroupTagHandler) happened in HtmlFormValidator, instead of in the tag handlers themselves, because we really only need to perform this logic a small fraction of the time (when an admin is editing a form schema, not when end users are entering/editing a form). Did you consider this at all?  I do realize that it would require some significant changes/improvements to HtmlFormValidator.

omod/src/main/java/org/openmrs/module/htmlformentry/web/controller/HtmlFormController.java
```diff
@@ -80,9 +81,16 @@ public void showHtmlForm(Model model,
...
                 if (fes.getFieldAccessorJavascript() != null) {
                 	html += "<script>" + fes.getFieldAccessorJavascript() + "</script>";
                 }
+				if (!formEntryContext.getInvalidObsGroupConceptIds().isEmpty()) {
+					model.addAttribute("invalidObsGroupConceptIds", formEntryContext.getInvalidObsGroupConceptIds());
+				}
+				if (!formEntryContext.getInvalidSetMemberConceptIds().isEmpty()) {
+					model.addAttribute("invalidSetMemberConceptIds", formEntryContext.getInvalidSetMemberConceptIds());
+				}
                 model.addAttribute("previewHtml", html);
```
mogoodrich: Shouldn't we also be testing when calling the saveHtmlForm method?

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
djazayeri: +1 to this comment. This should really be tested at time of saving the form definition. It is not useful to test this when a regular user is filling it out, and showing them a warning is only going to be confusing, because what can they do at that point?

geoff-wasilwa:
:) thanks, your comments really got me thinking. Working on moving the validations to HtmlFormValidator.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
geoff-wasilwa: What do you think of storing the error counter in HtmlFormValidator?

api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsTagHandlerTest.java
geoff-wasilwa: 'digDeeper' is a side effect of copy paste :)

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
mogoodrich: Having the HtmlFormValidator handle the validation makes sense... if using an error counter within it is the best way to go, then yes.  Not sure if moving to the validator will lead you to change your design complete though.

api-tests/src/test/java/org/openmrs/module/htmlformentry/handler/ObsTagHandlerTest.java
mogoodrich: Yeah, after posting that comment, I saw the digDeeper in the original code.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
djazayeri: If the warnings will be generated by HtmlGeneratorElement#generateHtml(FormEntryContext) then it makes sense to track warnings here in FormEntryContext, because this is really the "HTML Generation Context". I don't like hardcoding a holder for just this one specific validation. I would expect the property to be

    List<FormGenerationWarning>

and there can be

    class ObsGroupWithInvalidMember implementes FormGenerationWarning

However that means that the warnings will be generated every time the form is used, which could be an unnecessary performance hit.

My own inclination would be to do something like introducing a new sub-interface that gives more details about compiling (because we can't directly change HtmlGenerationElement for backwards-compatibility)

    interface AnalyzableHtmlGeneratorElement extends HtmlGeneratorElement {
        HtmlGenerationAnalysis analyze(FormEntryContext)
    }

geoff-wasilwa: Thanks @djazayeri, was struggling with the implementation of the warnings. Although I have changed the design, such that HtmlFormValidator creates a Document object from HtmlForm's xmlData and traverses the document validating obsgroup and obs element. 

I also had to change the stringToDocument method of HtmlFormEntryUtil so as to add line and column numbers as per this [blog](https://eyalsch.wordpress.com/2010/11/30/xml-dom-2/) and this [SO question](http://stackoverflow.com/questions/4915422/get-line-number-from-xml-node-java).

What's your take on this?

geoff-wasilwa: Aside: the initial plan was to have the warnings as a string collection in HtmlFormValidator

mogoodrich: Offhand, having the validator create a Document object from the xmlData makes sense to me--opens up the possibility of adding other document-based validations in the future.

I just glanced at the blog and question you linked to, but i assume you want to include line numbers so that you can point the end user to where the actual error is in the file?  That does sound helpful.  How are you planning on doing this?

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormValidator.java
```diff
@@ -66,4 +90,117 @@ private boolean hasEncounterTypeTag(String xml) throws Exception {
 		Node formNode = HtmlFormEntryUtil.findChild(doc, "htmlform");
 		return HtmlFormEntryUtil.findChild(formNode, "encounterType") != null;
 	}
+    
+	private void setHtmlFormWarnings(String warning) {
+		htmlFormWarnings.add(warning);
+	}
+	
+	private void traverseNode(Node currentNode, Node parentObsGroupNode, Errors errors) {
```
geoff-wasilwa: @mogoodrich this is what I had in mind, although I've been having internal arguments regarding implementation of the validators (it might grow out of control)

mogoodrich: @geoff-wasilwa looks like a good start to me. I agree that this could grow into a long if/then block as the number of validators grows, which would be ugly.  I'm not *too* concerned about it if you want to do it this way, but, yeah, if you could figure out a better lookup method for validators, that would be cool.  Potentially, even, each TagHandler could implement the validation for that tag within the class?  Then you could model something after the applyTagsHandler methods in HtmlFormEntryGenerator. Actually, maybe you could add a new interface TagValidator, that a TagHandler could optionally implement? (It would have to be optional at this point because there are TagHandlers defined outside of the HFE module that we wouldn't want to break).

Does that make sense? @djazayeri, let me know if you think I'm going down the wrong path.

djazayeri: Yes, I agree with a new subinterface of TagValidator (and that we have to
make this optional because of backwards-compatibility).

I was going to comment along these lines earlier. Maybe I did? Or maybe I
never posted that comment...

mogoodrich: Darius, looks like you did make the comment.  Do you think this sub-interface should be on the TagHandler or the HtmlGeneratorElement?

djazayeri: Offhand I'm not sure. I didn't look closely when saying HtmlGeneratorElement.

geoff-wasilwa: Okay guys, thanks for the comments let me work on the changes.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormValidator.java
```diff
@@ -16,6 +25,8 @@
 
 	protected final Log log = LogFactory.getLog(getClass());
 	
+	private List<String> htmlFormWarnings = new ArrayList<String>();
```
geoff-wasilwa: Will the string implementation of the warnings work or should I go with the earlier suggestion of having a `List` of `FormGenerationWarning`?

mogoodrich: Seems like it would be nice to have some sort of object that in addition to the string also includes the element tag, line number, etc... but I'd be fine with just string as well...

geoff-wasilwa:
Sorry I took long to deliver on this. Got a few issues though:
- can I use the .warn class found in uicommons module to represent the warnings?
- will I have to bind the warnings to request (`request.setAttribute(..)`) so that the warnings are displayed once the form has been saved?

geoff-wasilwa:
@mogoodrich and @djazayeri any thoughts?

geoff-wasilwa:
@mogoodrich I have made the changes. Thanks for pointing them out.

mogoodrich:
Great!  I will merge this is in now and then try to test it out on some of our forms.  Thanks!

PR #87, opened 2015-07-22
HTML-598 : Add Boolean datatype support to the <when></when> markup

rbuisson:
HTML-598 : Pull request to add the Boolean datatype support to the markup.
https://issues.openmrs.org/browse/HTML-598


api/src/main/java/org/openmrs/module/htmlformentry/handler/WhenTagHandler.java
```diff
@@ -38,17 +38,26 @@ protected String getSubstitution(FormEntrySession session, FormSubmissionControl
...
+			obs.whenValueThenDisplaySection(whenCondition, thenDisplay);
+		}
+		if (thenJavaScript != null) {
+			obs.whenValueThenJavaScript(whenCondition, thenJavaScript);
+		}
+		if (elseJavaScript != null) {
+			obs.whenValueElseJavaScript(whenCondition, elseJavaScript);
+		}
         return "";
     }
```
mogoodrich: Looks good.  Can you add a quick unit test as well? It wasn't obvious where the existing test for this tag was, but looks like it is in ObsTagTest:testObsWithControlsSection.  This is just a test that on the server-side the correct javascript is generated. Thanks!

rbuisson:
OK, thanks for the tip about this JUnit test. I would have never find it else.
Junit test ObsTagTest:testObsWithControlsSection is now updated.
(Hope I did it the right way)

api/src/main/java/org/openmrs/module/htmlformentry/handler/WhenTagHandler.java
rbuisson: See next commit

dkayiwa:
@mogoodrich do you think this is ready for merging? :)

mogoodrich:
Thanks for the reminder Daniel... I will merge this in.

PR #88, opened 2015-07-28
RA-865 fixed liquibase file

diospark:


api/src/main/resources/liquibase.xml
```diff
@@ -169,9 +169,12 @@
...
 
-        <sql>
-            UPDATE htmlformentry_html_form SET uuid = UUID() WHERE uuid is null;
-        </sql>
+        <customChange class="org.openmrs.module.htmlformentry.uuid.GenerateUuid">
+            <param name="tableNames">
+                htmlformentry_html_form
+            </param>
+            <param name="columnName" value="uuid"/>
+        </customChange>
```
mogoodrich: I believe this will fail because you can't change an old changeset unless you specify a new valid checksum. We will need to test this both "from scratch" and against an installation that already has HFE installed.

api/src/main/java/org/openmrs/module/htmlformentry/uuid/GenerateUuid.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Way to specify the table id columns that don't follow the table_name.table_name_id pattern
+     *
+     * @param idExceptions
+     */
+    public void setIdExceptions(String idExceptions) {
+        this.idExceptions = idExceptions;
+    }
+
+}
```
mogoodrich: Where did this come from?  Looks generic--are we using elsewhere within OpenMRS?  (Seems like, if it works, it should be our standard approach for this issue).

diospark: this is from openmrs-core
https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/util/databasechange/GenerateUuid.java

with updated method setUp()

mogoodrich: Cool, thanks.

diospark: I made a copy of this class, because it is not possible to override original one

mogoodrich: Did you not include it from core directly because that would require backporting the class to several earlier versions of OpenMRS?

mogoodrich: Seems fair... can you just add a comment to this effect in the class.

dkayiwa:
@diospark can you squash these commits into one? Did you also test this out on a new and existing database?

dkayiwa:
Am closing this for now until when it is fully tested.

PR #96, opened 2015-12-19
Fixed spelling errors and unclear wordings in /api and /api-2.0

alexianchen:


robbyoconnor:
@mogoodrich

api-2.0/src/main/java/org/openmrs/module/htmlformentry/PatientServiceCompatibility2_0.java
```diff
@@ -28,6 +28,6 @@
 
 	@Override
 	public void exitFromCare(Patient patient, Date dateExited, Concept reasonForExit) throws APIException {
-		//do nothing because this was moved to the exit from care module
+		//do nothing because this was moved to the exit from care module (remove?)
```
robbyoconnor: Why is `{remove?}` written here? Were you suggesting to remove this?

robbyoconnor:
Consider remove your formatting changes from this PR and rebasing and then making a separate PR for that?

api-2.0/src/main/java/org/openmrs/module/htmlformentry/PatientServiceCompatibility2_0.java
alexianchen: Yeah! it was "moved to the exit from care module", so shouldn't this be removed now as its redundant?

robbyoconnor:
You dont need to close this 

Also, use feature branches!!!!!

robbyoconnor:
Well actually you do -- you should use feature branches.

PR #119, opened 2016-12-29
Work In Progress - HTML-647 : Remove the restriction requiring the need for location and provider tags to be added to forms 

ssmusoke:


dkayiwa:
@ssmusoke Mark seems to suggest adding a required attribute with a default value of true. Did you intentionally ignore his suggested approach? Then one who does not want to require it can set to false. :)

ssmusoke:
@dkayiwa Thanks I had only tested the scenario where the tag is left out. Let me also add the not required attribute

dkayiwa:
@ssmusoke your current changes completely remove the validation, which i think is not the intention of the ticket. Someone who wants to enforce it should be able to do so. For instance those who run the module on lower platform versions that still require all these fields.

ssmusoke:
@dkayiwa Got you - I was thinking that the ticket looked at not having to add the fields fields to the forms. What you say makes sense, the fields are added but not required

api-tests/src/test/java/org/openmrs/module/htmlformentry/RegressionTest.java
```diff
@@ -96,6 +96,94 @@ public void testResults(SubmissionResults results) {
...
+			
+			@Override
+			public void setupRequest(MockHttpServletRequest request, Map<String, String> widgets) {
+				request.addParameter(widgets.get("Date:"), dateAsString(date));
+			}
+			
+			@Override
+			public void testResults(SubmissionResults results) {
+				results.assertNoErrors();
+				results.assertEncounterCreated();
```
mogoodrich: You might want to change this test to explicitly test that the location and provider have *not* been set on the created encounter.

mogoodrich:
@ssmusoke looks good, I'm going to merge this in... I've added one minor comment to the request, but there's a bigger comment I'll add to the original ticket itself for reference.

mogoodrich:
Actually, @ssmusoke I had to revert because the tests were failing. Can you take a look? 

Also, note that I think you'll need to modify the ProviderAndRoleElement to get it to work for the "encounterProviderAndRole" tag, which you'll probably want to add as well.

ssmusoke:
@mogoodrich The tests were failing and I was not sure why hence the WIP in the title. I will finish this up tomorrow and advise accordingly

mogoodrich:
Yeah, sorry, I noticed the WIP after I already merged... :)

PR #121, opened 2017-01-04
HTML-647 : Remove the restriction requiring the need for location and provider tags to be added to forms

ssmusoke:
@mogoodrich I have a strange failing test when I run 

`mvn test -Dtest=RegressionTest#testSimplestFormSuccessNoLocationAndNoProvider`

The failure is due to [w6] being null, yet this is an empty field error span in the HTML as 

`HTML <span id='encounterLocation'><select id="w5" name="w5"><option value="" selected="true">htmlformentry.chooseALocation</option><option value="1004">Boston</option><option value="1001">Kigali</option><option value="1006">Lacolline</option><option value="1002">Mirebalais</option><option value="1005">Scituate</option><option value="1">Unknown Location</option><option value="2">Xanadu</option></select><span class="error field-error" style="display: none" id="w6"></span></span>`

I have seen this with other test cases in the RegressionTest.java class but which does not cause failures


mogoodrich:
@ssmusoke  it looks like the validation is failing because no location has been specified.  Looks like it is not failing cleanly, but rather with a NPE, which is why we aren't seeing a helpful message, but the reason it is associated with w6 (instead of w5) it because it associates the message with the error widget tied to the widget that generated the error, not the widget itself, because this is where the message should be displayed on the form.

Looks like the test is failing because because EncounterDetailSubmissionElement:validationSubmission is failing on line 711 because it is not null safe.  (It should have been null safe in the first place and we missed this,and your change helped us discover the existing bug :).

(I suspect there will be similar problem with Provider as well).

ssmusoke:
@mogoodrich That helps, will keep working on it, now I know I am on the right track.

In the previous PR you said: Also, note that I think you'll need to modify the ProviderAndRoleElement to get it to work for the "encounterProviderAndRole" tag, which you'll probably want to add as well.

Where do I find the encounterProviderAndRole tag definitions? Is this in the htmlformentry19ext module?

mogoodrich:
The htmlformentryext19ext module content recently got moved into the main htmlformentry module, so I believe everything you need to modify should be in the htmlformentry module at this point.  (Starting with HFE 3.3.1, you should no longer need to have the HFE1.9Ext module installed--so it should be able to be dropped from the Ref App in the next release)

ssmusoke:
@mogoodrich Thanks will have more progress tomorrow

ssmusoke:
@mogoodrich The EncounterProviderAndRole tag already has tests for required attribute

mogoodrich:
Cool, thanks @ssmusoke... so we appear ready to go ahead and merge this?

ssmusoke:
@mogoodrich I think we are good to go here

mogoodrich:
Cool, merging now

PR #122, opened 2017-01-21
HTML-655 - added a location function to VelocityFunctions

ssmusoke:


api/src/main/java/org/openmrs/module/htmlformentry/VelocityFunctions.java
```diff
@@ -67,10 +70,32 @@ private AdministrationService getAdministrationService() {
...
+	 */
+	public Location location(String locationIdentifier) {
+		if (StringUtils.isNumeric(locationIdentifier)) {
+			return getLocationService().getLocation(new Integer(locationIdentifier));
+		}
+		Location location = getLocationService().getLocationByUuid(locationIdentifier);
+		if (location == null) {
+			return getLocationService().getLocation(locationIdentifier);
+		}
+		return location;
```
mogoodrich: I'm tried to centralize helper functions for getting various OpenMrs data and metadata in the HtmFormEntryUtils class.
It woudl make sense to just defer to HtmlFormEntryUtil.getLocation(String) because it should handle a few other formats as well.

ssmusoke: @mogoodrich I totally agree let me update

PR #130, opened 2018-03-08
fixed HTML-573

piyush9620:
fixed issue https://issues.openmrs.org/browse/HTML-573


piyush9620:
screen shot 2018-03-09 at 11 23 07 am

I tried checking, it seems to work. Here is an image from a form in french.

piyush9620:
There is also an existing instance of messages being fetched via jquery, at https://github.com/piyush9620/openmrs-module-htmlformentry/blob/25632838bac83a333ad732b9dd91d6299e5f117c/omod/src/main/webapp/resources/htmlFormEntry.js#L529
in the same file.

mogoodrich:
Great, thanks @piyush9620 !

mogoodrich:
Great, thanks @piyush9620 !

PR #133, opened 2018-03-10
HTML-366 Form Schema Preview should show a warning if an answer is not legal for its question

piyush9620:
https://issues.openmrs.org/browse/HTML-366

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
@@ -1,6 +1,7 @@
 package org.openmrs.module.htmlformentry.handler;
 
 import org.openmrs.Concept;
+import org.openmrs.ConceptAnswer;
 import org.openmrs.LocationTag;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.htmlformentry.BadFormDesignException;
@@ -19,6 +20,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.StringTokenizer;
+import java.util.Collection;
 
 /**
  * Handles the {@code <obs>} tag
@@ -92,6 +94,9 @@ public class ObsTagHandler extends AbstractTagHandler {
         List<Concept> concepts = new ArrayList<Concept>();
         String conceptId = HtmlFormEntryUtil.getNodeAttribute(node, "conceptId", null);
         String conceptIds = HtmlFormEntryUtil.getNodeAttribute(node, "conceptIds", null);
+        String answerConceptId = HtmlFormEntryUtil.getNodeAttribute(node, "answerConceptId", null);
+        String answerConceptIds = HtmlFormEntryUtil.getNodeAttribute(node, "answerConceptIds", null);
+	    
         if (conceptId == null && conceptIds == null) {
             analysis.addError(Context.getMessageSourceService().getMessage("htmlformentry.error.invalidConceptIdsAttribute",
                 new Object[] { "[concept id]", "[concept ids]" }, null));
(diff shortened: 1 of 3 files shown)


piyush9620:
@mogoodrich , i have tried to fix this ticket as follows:
created a list of valid answer ids depending on the question concepts.
created another list of actual answer ids that correspond to the answer concepts ids.
then compared if the actual concept ids are present in the valid concept ids or not.

mogoodrich:
Thanks @piyush9620, I will take a look, hopefully later today... just getting caught up on my Monday morning email now!

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -104,6 +113,11 @@ public TagAnalysis validate(Node node) {
             if (concept == null) {
                 analysis.addError(Context.getMessageSourceService().getMessage("htmlformentry.error.invalidConcept",
                     new Object[] { conceptId }, null));
+            } else {
+                Collection<ConceptAnswer> conceptAnswers = concept.getAnswers();
+                for (ConceptAnswer ca : conceptAnswers) {
```
mogoodrich: I think that you might run into a NPE exception here in some cases... some observations are numeric or free-text (you enter a number or a text value), so they don't have "answeConceptIds"... this should be null-safe, or perhaps you only need to do this if on of the "answer*" node attributes has been specified.

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -92,6 +94,13 @@ public TagAnalysis validate(Node node) {
         List<Concept> concepts = new ArrayList<Concept>();
         String conceptId = HtmlFormEntryUtil.getNodeAttribute(node, "conceptId", null);
         String conceptIds = HtmlFormEntryUtil.getNodeAttribute(node, "conceptIds", null);
+        String answerConceptId = HtmlFormEntryUtil.getNodeAttribute(node, "answerConceptId", null);
+        String answerConceptIds = HtmlFormEntryUtil.getNodeAttribute(node, "answerConceptIds", null);
+        String answerClasses = HtmlFormEntryUtil.getNodeAttribute(node, "answerClasses", null);
```
mogoodrich: Let's not worry about answerClasses for now, it's kind of a separate use case, and it's unlikely that single concept would have *all* elements of a single class in it's answer set.

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -92,6 +94,13 @@ public TagAnalysis validate(Node node) {
...
+        String answerConceptSetIds = HtmlFormEntryUtil.getNodeAttribute(node, "answerConceptSetIds", null);
```
mogoodrich: Let's skip answerConceptIds for now as well and just handle the base use case, when the end use specifies a "conceptId' and then either a "answerConceptId" or an "answerConceptIds".

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -112,10 +126,39 @@ public TagAnalysis validate(Node node) {
                 if (c == null) {
                     analysis.addError(Context.getMessageSourceService().getMessage("htmlformentry.error.invalidConcept",
                         new Object[] { conceptIdString }, null));
+                } else {
+                    Collection<ConceptAnswer> conceptAnswers = c.getAnswers();
```
mogoodrich: In the case of when the user specifics "conceptIds", it is used in conjunction with a single "answerConceptId", and what we actually need to test if that *each* of the concepts referenced has the answerConcept in it's answer set.  You can look at the Obs tag explanation in the HTML Form Entry Reference page for a greater understand of how this works. You can skip this use case now if you like or consider it an "extra credit" part of the ticket, but I don't think this is doing what we want it do here.

api/src/main/java/org/openmrs/module/htmlformentry/handler/ObsTagHandler.java
```diff
@@ -112,10 +126,39 @@ public TagAnalysis validate(Node node) {
...
+                    }
                 }
                 concepts.add(c);
             }
         }
+
+        if(answerConceptId != null) {
+            Concept c = HtmlFormEntryUtil.getConcept(answerConceptId);
+            actualAnswerConceptIDs.add(c.getConceptId());
+        } else if (answerConceptIds != null || answerConceptSetIds != null || answerClasses != null) {
```
mogoodrich: Per my comments above, let's just handle the answerConceptIds and answerConceptId use cases here, and ignore answerConceptSetIds and answerClasses for now.

piyush9620:
@mogoodrich can u plz check now.
Thanks.

piyush9620:
@mogoodrich  can u plz check now? 
Thanks.

Also about the tests that you asked me to write, i need a few pairs of valid question concept ids along with their valid answer concept ids. can u help me with these?

mogoodrich:
This looks good, thanks @piyush9620 !

For the tests, take a look at the "standardTestDataset.xml" in OpenMRS Core... this dataset is loaded into the in-memory test database before the tests are run.  I haven't checked fully, but I suspect it will have some question and answer concepts you can use.

If this isn't enough, you could potentially add your own test dataset, search for "executeDataSet" for an example of how that is used.

piyush9620:
@mogoodrich, added a few tests and left out others which where redundant.
Thanks.

mogoodrich:
Thanks!  Crazy day for me today, but should be able to review today or tomorrow...

mogoodrich:
Looks good, thanks @piyush9620 !

PR #134, opened 2018-03-16
HTML-325 Improve parser so that it can handle &nbsp; and other HTML-s…

piyush9620:
https://issues.openmrs.org/browse/HTML-325


piyush9620:
@mogoodrich, ca u plz review this. 
Thanks a ton.

PR #136, opened 2018-03-24
HTML-352 testEditEncounter never called in Regression Test Helper

dilekamadushan:
Issue Url -https://issues.openmrs.org/browse/HTML-352
deleted the testEditEncounter method


mogoodrich:
looks good, thanks!

PR #146, opened 2019-05-07
HTML-687 Should be possible to conditionally enroll a patient in a program

brandones:
https://issues.openmrs.org/browse/HTML-687

This adds an option to the EnrollInProgram tag called `showCheckbox`. If `showCheckbox="true"` is passed, a checkbox is rendered, and the patient is only enrolled in the program if the checkbox is checked. If the patient is already enrolled in the program, the checkbox is initialized checked and disabled. Support for disabling is added to CheckboxWidget.

### DIFF
--- a/api-tests/src/test/java/org/openmrs/module/htmlformentry/EnrollInProgramTagTest.java
+++ b/api-tests/src/test/java/org/openmrs/module/htmlformentry/EnrollInProgramTagTest.java
@@ -25,69 +25,68 @@ import org.springframework.mock.web.MockHttpServletRequest;
  * Tests of the various program-related tags
  */
 public class EnrollInProgramTagTest extends BaseModuleContextSensitiveTest {
-	
+
 	protected static final String XML_DATASET_PATH = "org/openmrs/module/htmlformentry/include/";
-	
+
 	protected static final String XML_REGRESSION_TEST_DATASET = "regressionTestDataSet";
-	
+
 	PatientService ps;
-	
+
 	ProgramWorkflowService pws;
-	
+
 	@Before
 	public void loadConcepts() throws Exception {
 		ps = Context.getPatientService();
 		pws = Context.getProgramWorkflowService();
 		executeDataSet(XML_DATASET_PATH + new TestUtil().getTestDatasetFilename(XML_REGRESSION_TEST_DATASET));
 	}
-	
+
 	@Test
 	public void enrollInProgram_shouldEnrollInProgramOnEncounterDate() throws Exception {
 		final Integer patientId = 2;
 		final Integer programId = 10;
 		//sanity check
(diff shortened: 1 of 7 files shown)


api-tests/src/test/java/org/openmrs/module/htmlformentry/EnrollInProgramTagTest.java
```diff
@@ -1,675 +1,860 @@
-package org.openmrs.module.htmlformentry;
```
brandones: I don't know why it's like this.

brandones:
@mogoodrich ?

api-tests/src/test/java/org/openmrs/module/htmlformentry/EnrollInProgramTagTest.java
HerbertYiga: @brandones   some times times untracked files during the  pushing make the code to appear as changed!!

api/src/main/java/org/openmrs/module/htmlformentry/element/EnrollInProgramElement.java
```diff
@@ -54,6 +62,24 @@ public EnrollInProgramElement(FormEntryContext context, Map<String, String> para
...
+
+		if ("true".equalsIgnoreCase(parameters.get("showCheckbox"))) {
+		    if (parameters.get("checkboxLabel") != null) {
+				checkToEnrollWidget = new CheckboxWidget(parameters.get("checkboxLabel"), "true");
+			} else {
+				checkToEnrollWidget = new CheckboxWidget();
+			}
+			// If patient is already enrolled, check and disable the checkbox
+			Patient patient = context.getExistingPatient();
+			List<PatientProgram> pps = Context.getProgramWorkflowService().getPatientPrograms(patient, program, null, null, null, null, false);
```
mogoodrich: So this seems to *not* take date into account as all?  What if the patient was enrolled, but years ago, and is no longer enrolled? If you look at the enrollInProgram action (in form submission actions) you'll see that there's a "HtmlFormEntryUtil.isEnrolledInProgramOnDate" method that you could use to test to see if the patient is enrolled in the program on a specified date... the issue here, however, is that this is server-side, so you could run it with the current date on page load, but what happens if the user changes the encounter date or the enrollment date when editing the form?  This is why triggering/changing program enrollment via HFE can be tricky... we won't get things perfect here, but I wonder if just comparing against the current date is "good enough" and we can document the behavior in the docs.

brandones: Great catch, thanks Mark! I wrote a test to make sure it doesn't disable the checkbox for completed programs and updated the logic accordingly.

PR #147, opened 2019-05-13
HTML-690 enrollInProgram tag checkbox should be able to toggle show/hide elements

brandones:
https://issues.openmrs.org/browse/HTML-690

Now that the enrollInProgram tag can create a checkbox that will conditionally enroll a patient, it would be great if that checkbox could toggle whether to show or hide a section, like the Obs checkboxes.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/EnrollInProgramElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/EnrollInProgramElement.java
@@ -29,6 +29,7 @@ import org.openmrs.module.htmlformentry.action.FormSubmissionControllerAction;
 import org.openmrs.module.htmlformentry.widget.CheckboxWidget;
 import org.openmrs.module.htmlformentry.widget.DateWidget;
 import org.openmrs.module.htmlformentry.widget.ErrorWidget;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/EnrollInProgramElement.java
```diff
@@ -81,6 +82,16 @@ public EnrollInProgramElement(FormEntryContext context, Map<String, String> para
...
 		}
+
+		String toggleParameter = parameters.get("toggle");
+		if (toggleParameter != null) {
+		    if (checkToEnrollWidget == null) {
+				throw new RuntimeException("<enrollInProgram> 'toggle' parameter requires 'showCheckbox=\"true\"'");
+			}
+			ToggleWidget toggleWidget = new ToggleWidget(toggleParameter);
+			checkToEnrollWidget.setToggleTarget(toggleWidget.getTargetId());
+			checkToEnrollWidget.setToggleDimInd(toggleWidget.isToggleDim());
```
mogoodrich: Seems weird that the toggle widget itself is never used itself, but I do see that this is the way it's used in ObsSubmission element...

brandones:
Yeah, it’s pretty weird... I also couldn’t figure out how to test it (it seems no one has) 😬

PR #148, opened 2019-06-13
HTML-696 - Improve performance of loading forms that utilize concept …

mseaton:
…mapping codes

This introduces a simple Map<String, Integer> cache on htmlformentryservice that maintains a cache of a mapping code in "source:code" format -> conceptId.  This enables much faster performance when loading an htmlform that contains the same concept mapping lookups multiple times (eg. in a repeat tag, or with common answers).  It also enables faster performance when loading forms subsequent to the first time a form is loaded after a context refresh.  An additional service method has been added to clear the cache, so distribution modules can choose to do this if needed upon activation if they load concepts via liquibase or SQL.


mogoodrich:
Sweet, thanks @mseaton ! LGTM

PR #149, opened 2019-06-25
HTML-622: Enable support for providerRoles when autocompleteProvider …

mseaton:
…is set to true

There is a fair amount of refactoring in this pull request, as I encountered several bugs along the way, and my overarching goal beyond fixing this ticket was to do so in order to improve performance of loading htmlforms, so some of the refactoring was done in order to ensure that this tag was able to load quickly, particularly when in the autocompleteProvider mode.

Much of the refactoring centers around moving the tag parsing out of the Element and into a new Tag class.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,3 @@
 language: java
 jdk:
- - oraclejdk8
+ - openjdk8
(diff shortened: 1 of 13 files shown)


HerbertYiga:
Some failing builds here!!

PR #155, opened 2019-08-19
HTML-705: Condition tag support from Core 2.2.0 and beyond.

samuelmale:
Followup commit on => https://issues.openmrs.org/browse/HTML-705


samuelmale:
Yes, it works as expected @mks-d

api-2.2/src/main/java/org/openmrs/module/htmlformentry/handler/ConditionTagHandlerSupport2_2.java
```diff
@@ -7,7 +7,7 @@
 import org.openmrs.module.htmlformentry.FormEntrySession;
 import org.openmrs.module.htmlformentry.FormSubmissionController;
 
-@OpenmrsProfile(openmrsPlatformVersion = "2.2.*")
+@OpenmrsProfile(openmrsPlatformVersion = "2.2.0")
```
mogoodrich: @samuelmale  @mks-d  what is the purpose behind this change?

PR #166, opened 2020-03-10
HTML-718, display retired drugs as greyed-out options

cioan:

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/RemoteJsonAutocompleteWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/RemoteJsonAutocompleteWidget.java
@@ -32,7 +32,7 @@ public class RemoteJsonAutocompleteWidget implements Widget {
     public void setInitialValue(Object initialValue) {
         if (initialValue instanceof Drug) {
             Drug asDrug = (Drug) initialValue;
-            this.initialValue = new Option(asDrug.getName(), "Drug:" + asDrug.getId(), true);
+            this.initialValue = new Option(asDrug.getName(), "Drug:" + asDrug.getId(), true, asDrug.getRetired());
         }
         else if (initialValue instanceof Concept) {
             Concept asConcept = (Concept) initialValue;
@@ -51,6 +51,9 @@ public class RemoteJsonAutocompleteWidget implements Widget {
         String formFieldName = context.getFieldName(this);
         StringBuilder sb = new StringBuilder();
         sb.append("<input id=\"" + formFieldName + "-display\"");
+        if (initialValue != null && initialValue.isRetired()) {
+            sb.append(" style=\"color: gray; font-style: italic\"");
+        }
         if (initialValue != null) {
             sb.append(" value=\"" + initialValue.getLabel() + "\"");
         }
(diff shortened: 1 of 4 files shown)


cioan:
@mogoodrich and @mseaton , here is the PR I mentioned to you in this morning. Thanks for reviewing it.

api/src/main/java/org/openmrs/module/htmlformentry/widget/RemoteJsonAutocompleteWidget.java
```diff
@@ -32,7 +32,7 @@ public void setDisplayTemplate(String displayTemplate) {
     public void setInitialValue(Object initialValue) {
         if (initialValue instanceof Drug) {
             Drug asDrug = (Drug) initialValue;
-            this.initialValue = new Option(asDrug.getName(), "Drug:" + asDrug.getId(), true);
+            this.initialValue = new Option(asDrug.getName(), "Drug:" + asDrug.getId(), true, asDrug.getRetired());
         }
         else if (initialValue instanceof Concept) {
             Concept asConcept = (Concept) initialValue;
```
mogoodrich: I'm wondering if we want to update the concept initialization here as well?  Do we know if this will work with concepts as well? (If it doesn't, no need to implement it yet, but if it's close, worth finishing off)

cioan: Thanks @mogoodrich . This makes sense. But, how do I test this with concepts? What kind of  form content would reach this code? The Drugs section is reached when the form contains an obs similar to this:
`<obs id="name{0}" class="medication-name" conceptId="PIH:1282" answerDrugs="true"/>`

mogoodrich: I'm not sure... I know there's support for a concept-search autocomplete in HFE, I was assuming it used this code, but it's possible it doesn't?

cioan: The only place I could find where this widget is used is here:
https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java#L775

and the above is triggered only when **answerDrugs=true**. Where is the concept-search autocomplete in HFE?

mogoodrich: Hmm, it's certainly possible it isn't used.  Check the HTML Form Entry Reference page for "autocomplete"?

https://wiki.openmrs.org/display/docs/HTML+Form+Entry+Module+Reference

cioan: The concept-search autocomplete does not use the RemoteJsonAutocompleteWidget:
https://github.com/openmrs/openmrs-module-htmlformentry/blob/05928f63b5936b68bf98a32e176870cb16c827e7/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java#L769

mogoodrich:
Okay, maybe it isn't used then!

PR #168, opened 2020-03-25
HTML-717: Fetching complex obs RAW_VIEW with ComplexObsServlet.

mkumar214:
Wrong URL is generating for download the image which is saved in complex concept

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/WidgetFactory.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/WidgetFactory.java
@@ -175,10 +175,10 @@ public class WidgetFactory {
 		if (ModuleUtil.compareVersion(OpenmrsConstants.OPENMRS_VERSION_SHORT, "2.0") < 0) {
 			prefix = "";
 		}
-		return "/" + WebConstants.WEBAPP_NAME + prefix + "/complexObsServlet?obsId=" + obs.getObsId();
+		return "/" + WebConstants.WEBAPP_NAME + prefix + "/complexObsServlet?obsId=" + obs.getObsId() + "&view=RAW_VIEW";
 	}
 	
 	private static String getDownloadHyperlink(Obs obs) {
-		return getViewHyperlink(obs) + "&view=download&viewType=download";
+		return getViewHyperlink(obs) + "&download=view";
 	}
 }


api/src/main/java/org/openmrs/module/htmlformentry/widget/WidgetFactory.java
```diff
@@ -175,10 +175,10 @@ private static String getViewHyperlink(Obs obs) {
 		if (ModuleUtil.compareVersion(OpenmrsConstants.OPENMRS_VERSION_SHORT, "2.0") < 0) {
 			prefix = "";
 		}
-		return "/" + WebConstants.WEBAPP_NAME + prefix + "/complexObsServlet?obsId=" + obs.getObsId();
+		return "/" + WebConstants.WEBAPP_NAME + prefix + "/complexObsServlet?obsId=" + obs.getObsId() + "&view=RAW_VIEW";
 	}
 	
 	private static String getDownloadHyperlink(Obs obs) {
-		return getViewHyperlink(obs) + "&view=download&viewType=download";
```
dkayiwa: Isn't safe to leave the original params and just add the news ones?

chitralekhak: > Isn't safe to leave the original params and just add the news ones?
@dkayiwa , the ComplexObsServlet in openmrs-module-legacyui expects 3 parameters: obsId, view ,download
Also the view is available only if the view attribute is "RAW_VIEW" -  refer to getObs() in -ImageHandler in openmrs-api

rrameshbtech: @dkayiwa Could we review this PR? As mentioned by @chitralekhak above, the change looks fine. Also we have checked in our context it works fine for our scenario.

mks-d: @rrameshbtech and @chitralekhak, why was it ok to scrap the `viewType` param?

@dkayiwa I think it's ok to default to the `RAW_VIEW` view on the hand, but I don't think you were concerned either by that, correct?

rrameshbtech: @mks-d We are able to see only view, obsid & download are read as parameter in below class.
https://github.com/openmrs/openmrs-module-legacyui/blob/d97268394738d3bad95674b823e538d732ac886f/omod/src/main/java/org/openmrs/web/servlet/ComplexObsServlet.java#L45

I think @dkayiwa question is about reusing viewHyperlink to create download Hyperlink.

dkayiwa: Am scared of breaking whoever was relying on the old parameters. That is why i was thinking of simply appending the new ones.

mks-d: @dkayiwa `viewType` is not even a supported param, see [here](https://github.com/openmrs/openmrs-module-legacyui/blob/d97268394738d3bad95674b823e538d732ac886f/omod/src/main/java/org/openmrs/web/servlet/ComplexObsServlet.java#L45-L47https://github.com/openmrs/openmrs-module-legacyui/blob/d97268394738d3bad95674b823e538d732ac886f/omod/src/main/java/org/openmrs/web/servlet/ComplexObsServlet.java#L45-L47). As of for `view=download`... I have no idea what that was about. It almost feels like a mistake this:
```java
"&view=download&viewType=download"
```
As in, that both params were set to `download`.

@mogoodrich, any idea?

_P.S._ However there is a `download` _param_... So basically this seems like a valid option:
```java
"&view=RAW_VIEW&download=true"
```

mogoodrich: Hmmm... sorry, unfortunately not sure of the history here @mks-d  and I don't believe we use the complex obs handler in any of our forms so we may never have used/tested this at PIH EMR.  It's entirely possible that this was a mistake.  But  worth a quick check in the old ComplexObsServlet in Core 1.x if we haven't checked already?

lnball:
@mogoodrich @mseaton I think we used complex obs (once upon a time on the HIREF Peru project for xrays), but not on PIH EMR.  And not used now.

mks-d:
I'm inclined to let this go through and let people raise their voices **if** we happen to break something. We will not have the room to test against 1.x tho @mogoodrich, there is no way to justify that on our end.

mogoodrich:
@mks-d I'm not suggesting testing, just taking a look at the ComplexObsServlet in Core in like the 1.9 or 1.10 branch and see if view or viewType is used at all.

api/src/main/java/org/openmrs/module/htmlformentry/widget/WidgetFactory.java
mks-d: Sure @mogoodrich ;-)
Yes `viewType` [did exist in 1.x](https://github.com/openmrs/openmrs-core/blob/40e13ad07a9f66e483c0ac6e19924376e1028bc6/web/src/main/java/org/openmrs/web/servlet/ComplexObsServlet.java#L47).

Ok so, @rrameshbtech @chitralekhak let's do this then:
```java
"&view=RAW_VIEW&download=true&viewType=download"
```

It doesn't feel we should give too much space to `view=download` though, that one feels wrong.

rrameshbtech: Thanks @mks-d. As we do not have access to this fork, we have created a new PR with above mentioned change.

Please review the same. 
https://github.com/openmrs/openmrs-module-htmlformentry/pull/175/files

mks-d:
Replaced by https://github.com/openmrs/openmrs-module-htmlformentry/pull/175.

PR #169, opened 2020-03-27
HTML-721 : retroactive entry updates patient program enrollment date

kml27:
update patient program enrollment date if the state startDate is before the current PatientProgram dateEnrolled

HTML-721 Retroactively adding a state to a patient program before the current enrollment date should update the PatientProgram enrollment Date

When a program workflow state is retroactively entered for an encounterDatetime before the enrollment of the existing program, and a new program workflow state is entered, the enrollment date is updated on a new copy, and the original PatientProgram is voided to leave an audit trail.

Additional assertions were added in the existing unit tests where this was observed. 

## Issue I worked on

see https://issues.openmrs.org/browse/HTML-721
this is a prereq for https://issues.openmrs.org/browse/HTML-720

- [x] My pull request only contains **ONE single commit**

(additional assertions anyway)

  No? -> execute above command

- [x] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
@@ -643,23 +643,7 @@ public class FormEntrySession {
 
         if (submissionActions.getPatientProgramsToUpdate() != null) {
             for (PatientProgram patientProgram : submissionActions.getPatientProgramsToUpdate()) {
-				
-				if (encounter.getEncounterDatetime().before(patientProgram.getDateEnrolled())) {
-					PatientProgram updatedProgram = patientProgram.copy();
-					
-					//voided now
-					patientProgram.setVoided(true);
-					patientProgram.setDateVoided(new Date());
-					patientProgram.setVoidReason("Retroactive entry update in encounter " + encounter.getUuid());
-					patientProgram.setVoidedBy(encounter.getCreator());
-					
-					patientProgram = updatedProgram;
-					
-					patientProgram.setDateEnrolled(encounter.getEncounterDatetime());
-            	}
-            	
-            	Context.getProgramWorkflowService().savePatientProgram(patientProgram);
-				
+                Context.getProgramWorkflowService().savePatientProgram(patientProgram);
             }
         }
 
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
@@ -643,7 +643,23 @@ public class FormEntrySession {
 
         if (submissionActions.getPatientProgramsToUpdate() != null) {
             for (PatientProgram patientProgram : submissionActions.getPatientProgramsToUpdate()) {
-                Context.getProgramWorkflowService().savePatientProgram(patientProgram);
+				
+				if (encounter.getEncounterDatetime().before(patientProgram.getDateEnrolled())) {
+					PatientProgram updatedProgram = patientProgram.copy();
+					
+					//voided now
+					patientProgram.setVoided(true);
+					patientProgram.setDateVoided(new Date());
+					patientProgram.setVoidReason("Retroactive entry update in encounter " + encounter.getUuid());
+					patientProgram.setVoidedBy(encounter.getCreator());
+					
+					patientProgram = updatedProgram;
+					
+					patientProgram.setDateEnrolled(encounter.getEncounterDatetime());
+            	}
+            	
+            	Context.getProgramWorkflowService().savePatientProgram(patientProgram);
+				
             }
         }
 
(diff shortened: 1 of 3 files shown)


kml27:
@mogoodrich

kml27:
@mseaton @mks-d Possibly review and approve if appropriate. HTML-720 is dependent on HTML-721. Thanks.

mogoodrich:
Sorry, this has been on my list to review since last week and it keeps getting pushed back, but I think I should get to this today.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -643,7 +643,23 @@ else if (outcomeObs.size() > 1) {
...
+					patientProgram.setVoidReason("Retroactive entry update in encounter " + encounter.getUuid());
+					patientProgram.setVoidedBy(encounter.getCreator());
+					
+					patientProgram = updatedProgram;
+					
+					patientProgram.setDateEnrolled(encounter.getEncounterDatetime());
+            	}
+            	
+            	Context.getProgramWorkflowService().savePatientProgram(patientProgram);
+
```
mogoodrich, line 662: Is there a reason we can't perform this logic in the handleSubmission method of the WorkflowStateSubmissoinElement?  It would be better to encapsulate all the logic about changes made to the program and state there.

mogoodrich: Specifically, it looks like there is already some logic there to set the program enrollment earlier if needed:

https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/element/WorkflowStateSubmissionElement.java#L286

I haven't looked into this code deeply but I assume it is buggy or not covering our use case?  The logic should be improved here if at all possible.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -643,7 +643,23 @@ else if (outcomeObs.size() > 1) {
...
-                Context.getProgramWorkflowService().savePatientProgram(patientProgram);
+				
+				if (encounter.getEncounterDatetime().before(patientProgram.getDateEnrolled())) {
+					PatientProgram updatedProgram = patientProgram.copy();
+					
+					//voided now
+					patientProgram.setVoided(true);
+					patientProgram.setDateVoided(new Date());
+					patientProgram.setVoidReason("Retroactive entry update in encounter " + encounter.getUuid());
+					patientProgram.setVoidedBy(encounter.getCreator());
```
mogoodrich, line 654: We can just update the enrollment date here, no need to void and recreate... there should be handlers behind-the-scenes to do that if necessary.  (Not sure if it really is in this case, though)

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
kml27: Ok, it does look like the comment was stating that was intended to change it to the previous date. I will look into why that wasn't already working. While I can see the design choice to have everything encapsulated in this class, I'd expect any custom handler/element that may add a new workflow state to have the same resulting change to the PatientProgram. (I was even considering making this a TRUNK ticket originally).

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
kml27: I was wondering about this too. I thought there was something in the hibernate hooks that might automate this sort of thing, but 1) "Reason" is one that a hook like that might not be able to provide a very meaningful message for, 2) How does it handle knowing when to use "Changed" update vs. duplicate and void, 3) there is code in the workflow state element that does set these too https://github.com/openmrs/openmrs-module-htmlformentry/blob/761a3b1c1d57be0fdd158092f8c7cf2d9ab5b1b7/api/src/main/java/org/openmrs/module/htmlformentry/element/WorkflowStateSubmissionElement.java#L255-L257

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
kml27: Since the states in question occur before the current program enrollment, the logic in FormSubmissionAction.transitionToState() https://github.com/openmrs/openmrs-module-htmlformentry/blob/761a3b1c1d57be0fdd158092f8c7cf2d9ab5b1b7/api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionActions.java#L477-L590 is finding the matching patientProgram to add the new state to, meaning the element code would need to access programsToUpdate() after that in order to move the logic from FormEntrySession to WorkflowStateSubmissionElement.

PR #170, opened 2020-04-02
HTML-713: New restrictToSupportedVisitLocations attribute to EncounterLocation tag.

HerbertYiga:
Ticket ID:https://issues.openmrs.org/browse/HTML-713

Description

Added new "restrictToSupportedVisitLocations" attribute to EncounterLocation tag


HerbertYiga:
@dkayiwa @mks-d  @mogoodrich  @ibacher added restrictToSupportedVisitLocations attribute to the  EncounterLocation tag.and it works as follows:
Locations of the available visits are returned when true,and when false,the default locations are rendered and when  not true or false,a run time error is thrown.I have added screen shorts on the ticket id.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -353,6 +354,23 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
             if (parameters.get("tags") != null && parameters.get("orders") != null) {
                 throw new RuntimeException("Using both \"order\" and \"tags\" attribute in an encounterLocation tag is not currently supported");
             }
+            if(parameters.get("restrictToSupportedVisitLocations")!=null) {
+            if ("true".equals(parameters.get("restrictToSupportedVisitLocations"))) {
```
mogoodrich, line 358: are we missing an indent here?

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -353,6 +354,23 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
...
+                List<Visit> visits=Context.getVisitService().getAllVisits();
```
mogoodrich, line 359: This will be *very* computationally expensive... rather than doing this, you should fetch locations that have the "Visit Location" tag and then add in all their child and grandchildren.

HerbertYiga:
@mogoodrich @dkayiwa @mks-d  made some changes and updated the screen shorts on the ticket id

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -353,6 +354,27 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
             if (parameters.get("tags") != null && parameters.get("orders") != null) {
                 throw new RuntimeException("Using both \"order\" and \"tags\" attribute in an encounterLocation tag is not currently supported");
             }
+            if(parameters.get("restrictToSupportedVisitLocations")!=null) {
+            	Set<Location>childVisitLocations=new HashSet<Location>();
+               if ("true".equals(parameters.get("restrictToSupportedVisitLocations"))) {
+            	  LocationTag visitLocationTag = HtmlFormEntryUtil.getLocationTag("Visit Location");
+              	  List<Location> visitLocations=Context.getLocationService().getLocationsByTag(visitLocationTag);
+              	 for(Location visitLocation:visitLocations) {
```
mogoodrich, line 362: Thanks @HerbertYiga... this is much less computational expensive!  

But I think you'll need to create a recursive function here, because not only do you want all child locations of a visit location, but then all child locations of those child locations and so forth.  You'll likely need a utility method to do this.

HerbertYiga:
@mogoodrich  @ibacher  @dkayiwa  @mks-d 

introduced a private recursive method called getAllChildAndGrandLocations that helps pick all child and grand locations.
On the ticket ID,i added the Location Hierarchy screen short plus the screen short of the out put after introducing the attribute

HerbertYiga:
@mogoodrich  helo,just a gentle reminder to have this reviewed.Thanks

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -835,5 +854,23 @@ private boolean hasTimeComponent(Date date) {
     private DateMidnight stripTimeComponent(Date date) {
         return new DateMidnight(date);
     }
+    
+    private Set<Location> getAllChildAndGrandLocations(Set<Location> visitLocations){
```
mogoodrich, line 858: Can you write a quick test to make sure this worked properly?  I may be wrong, but I'm not sure the accumulation of results will work correctly?

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -852,4 +871,27 @@ private DateMidnight stripTimeComponent(Date date) {
...
 	
+	public static Set<Location> getAllChildAndGrandLocations(Set<Location> visitLocations) {
+
+		Set<Location> allchildAndGrandVisitLocations = new HashSet<Location>();
+		for (Location visitLocation : visitLocations) {
+
+			if (visitLocation.getChildLocations() != null) {
+
+				allchildAndGrandVisitLocations.addAll(visitLocation.getChildLocations());
+				getAllChildAndGrandLocations(allchildAndGrandVisitLocations);
```
ibacher, line 882: The implementation here should probably be something like this:

```java
public static Set<Location> getLocationHierarchy(Set<Location> visitLocations) {
    return getChildLocations(visitLocations, new HashSet<>());
}

public static Set<Location> getLocationHierarchy(Set<Location> visitLocations, Set<Location> traversedLocations) {
    Set<Location> locations = new HashSet<>();
    for (Location visitLocation : visitLocations) {
        if (!traversedLocations.contains(visitLocation)) {
            locations.add(visitLocation);
            traversedLocations.add(visitLocation);
            
            if (visitLocation.getChildLocations() != null) {
                locations.addAll(getLocationHierarchy(visitLocation.getChildLocations(), traversedLocations));
            }
        }
    }

    return locations;
}
```

This helps with the case where a Visit Location is contained inside another Visit Location.

PR #171, opened 2020-04-04
HTML-722 : Encounter edit should not overwrite "unrelated" workflow state

kml27:
Setting a new patient program workflow state while editing an encounter should not overwrite previous states not related to the encounter

HTML-722: Setting a new program workflow state when editing an existing encounter overwrites previous state

https://issues.openmrs.org/browse/HTML-722

In <workflowState/> element impl. in handleSubmission during EDIT mode context, there must be an exact datetime match of the encounter (or previous if it was changed during the edit) with the start date of the most recent workflow state.

## Issue I worked on

see https://issues.openmrs.org/browse/HTML-722

- [x] My pull request only contains **ONE single commit**

  No? -> execute above command

- [x] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
+++ b/api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
@@ -1624,6 +1624,94 @@ public class WorkflowStateTagTest extends BaseModuleContextSensitiveTest {
 	}
 	
 	@Test
+	public void shouldNotOverwritePreviousStateOnEdit() throws Exception {
+		//Given: Patient has a workflow state of X starting in Jan 2012
+		transitionToState(START_STATE, PAST_DATE);
+		
+		new RegressionTestHelper() {
+			
+			@Override
+			public String getFormName() {
+				return XML_FORM_NAME;
+			}
+			
+			@Override
+			public String[] widgetLabels() {
+				return new String[] { "Date:", "Location:", "Provider:", "State:" };
+			}
+			
+			@Override
+			public void setupRequest(MockHttpServletRequest request, Map<String, String> widgets) {
+				request.addParameter(widgets.get("Location:"), "2");
+				request.addParameter(widgets.get("Provider:"), "502");
+				
+				//When: Html form is being back entered with an encounter date of June 2011, in which the workflow state selected is Y
+				request.addParameter(widgets.get("Date:"), dateAsString(DATE));
+				//request.addParameter(widgets.get("State:"), START_STATE);
+			}
+			
+			@Override
+			public void testResults(SubmissionResults results) {
+				results.assertNoErrors();
+				results.assertEncounterCreated();
+				results.assertProvider(502);
+				results.assertLocation(2);
+			}
+			
+			@Override
+			public boolean doEditEncounter() {
+				return true;
+			}
+			
+			@Override
+			public String[] widgetLabelsForEdit() {
+				return new String[] { "Date:", "Location:", "Provider:", "State:" };
+			}
+			
+			@Override
+			public void setupEditRequest(MockHttpServletRequest request, Map<String, String> widgets) {
+				request.setParameter(widgets.get("Location:"), "2");
+				request.setParameter(widgets.get("Provider:"), "502");
+				request.setParameter(widgets.get("Date:"), dateAsString(DATE));
+				request.setParameter(widgets.get("State:"), MIDDLE_STATE);
+			}
+			
+			@Override
+			public void testEditedResults(SubmissionResults results) {
+				results.assertNoErrors();
+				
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/WorkflowStateSubmissionElement.java
```diff
@@ -215,9 +215,21 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
...
+				}
+						
 				ProgramWorkflowState newState = Context.getProgramWorkflowService().getStateByUuid(stateUuid);
-				PatientState oldPatientState = getActivePatientState(session.getContext().getExistingPatient(), session
-				        .getContext().getPreviousEncounterDate(), workflow);
+				PatientState oldPatientState = getActivePatientState(session.getContext().getExistingPatient(), prevDate, workflow);
+
+				//since there is no direct connection between encounters and states
+				//the best that can be done, for now, is to check 
+				//if the encounterDatetime matches the startDate
```
mogoodrich, line 229: Trying to remember back all those years ago when I worked on this... this whole logic is rather muddled unfortunately...

I think I'm following your logic, but want to make sure: in this case "oldPatientState" means "the old patient state we assume was created via this form"?  And if we find an "old state we assume was created via the form" we want to *replace* that state with the new state... but if we find a state for on the previous encounter date, but it wasn't *started* on the previous encounter date, we assume it wasn't created by the form as so we don't want to *replace* the existing state, but instead just *transition* to the new state... is this correct?

mogoodrich: Also, can you confirm that you agree with the logic in this ticket is not  bug: https://tickets.openmrs.org/browse/HTML-321  (If you feel it is a bug we can debate, but just want to make sure we aren't changing assumptins or behavior without realizing it.

kml27: In response to your first comment, yes.

Re: HTML-321
I think the behavior of changing the start date to the encounter date when submitting an edit regardless of whether the state was "edited" is unexpected. The behavior of <workflowState/> when in enter mode when 1) submitting the "blank" option and 2) when submitting the same state as the current state, is to "do nothing". I believe "do nothing" would be the expected behavior, as it seems the reporter of the ticket did. 

Primarily, in HTML-321 the encounter datetime was not changed during edit. In my reading of the Notes section, this would fall under "Otherwise, the state selected on the form is compared to the current patient state at the time of the original encounter date. If these two states differ, change the existing patient state on the original encounter date is changed to the selected state." Since 1) there is an existing state, 2) there is no "new" encounter datetime to be before or after the "original", it is still the "original" encounter datetime.

mogoodrich: I think I agree @kml27 ... do your tickets change this behavior, or is this not affected for now?

kml27: I believe this still follows the existing spec as written. It's just a matter of if/which state will be edited. Having a full set of unit tests (if there are any that haven't already been implemented) to verify the described behaviors is probably the best way to answer the question though.

kml27:
This issue can't be completely resolved since program workflow states do not have a time of day component, even though encounterDatetime can (though its time component is often cleared). The time of day component could be cleared before the compare to widen the window for matches. Unfortunately, the opposite isn't quite possible right now. Created datetime for the encounter and state are not based on a single db transaction datetime, so they cannot reliably be used to match the encounter and state.

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
```diff
@@ -1623,6 +1623,94 @@ public void testBlankFormHtml(String html) {
...
+				//verify existing behavior does match expectations for the second state
+				PatientProgram patientProgram = getPatientProgramByState(results.getPatient(), secondState, DATE);
+				PatientState currentPatientState = getPatientState(patientProgram, secondState, DATE);
+				Assert.assertNotNull(patientProgram);
+				Assert.assertEquals(dateAsString(PAST_DATE), dateAsString(patientProgram.getDateEnrolled()));
+				Assert.assertEquals(dateAsString(DATE), dateAsString(currentPatientState.getStartDate()));
+				Assert.assertNull(patientProgram.getDateCompleted());
+				Assert.assertNull(currentPatientState.getEndDate());
+				
+				//illustrate that this second state has overwritten the existing state
```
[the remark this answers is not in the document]
mogoodrich: Update or remove, whatever works for you!

PR #174, opened 2020-04-16
HTML-712: <drugOrder/> tag to be supported on  platform 2.*

Ruhanga:
Ticket: https://issues.openmrs.org/browse/HTML-712

### DIFF
--- a/api-1.10/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement1_10.java
+++ b/api-1.10/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement1_10.java
@@ -417,7 +417,7 @@ public class DrugOrderSubmissionElement1_10 extends DrugOrderSubmissionElement {
 		drugOrder.setCareSetting(Context.getOrderService().getCareSetting(orderTag.careSettingId));
 		OrderFrequency orderFrequency = Context.getOrderService().getOrderFrequency(Integer.valueOf(orderTag.frequency));
 		drugOrder.setFrequency(orderFrequency);
-		drugOrder.setDateActivated(orderTag.startDate);
+		drugOrder.setDateActivated(session.getEncounter().getEncounterDatetime());
         drugOrder.setNumRefills(orderTag.numRefills);
 		//order duration:
 		if (orderTag.orderDuration != null)
--- a/api-2.0/pom.xml
+++ b/api-2.0/pom.xml
@@ -23,13 +23,23 @@
             <version>${project.parent.version}</version>
             <scope>provided</scope>
         </dependency>
-
+		<dependency>
+            <groupId>${project.parent.groupId}</groupId>
+            <artifactId>${project.parent.artifactId}-api-1.10</artifactId>
+            <version>${project.parent.version}</version>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>org.openmrs.api</groupId>
             <artifactId>openmrs-api</artifactId>
             <version>${openMRSVersion}</version>
             <scope>provided</scope>
         </dependency>
+        
+        <dependency>
+			<groupId>org.openmrs.web</groupId>
+			<artifactId>openmrs-web</artifactId>
+		</dependency>
 
         <dependency>
             <groupId>org.openmrs.api</groupId>
--- a/api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
+++ b/api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
@@ -1,0 +1,23 @@
+package org.openmrs.module.htmlformentry.element;
+
+import java.util.Date;
+import java.util.Map;
+import org.openmrs.DrugOrder;
+import org.openmrs.module.htmlformentry.FormEntryContext;
+
+public class DrugOrderSubmissionElement2_0 extends DrugOrderSubmissionElement1_10 {	
+	
+	public DrugOrderSubmissionElement2_0(FormEntryContext context, Map<String, String> parameters) {
+		super(context, parameters);
+	}
+	
+	@Override
+	protected Date getDrugOrderStartDate(DrugOrder dod) {
+		return dod.getDateActivated();
+	}
+	
+	@Override
+	protected void setOrderTagDose(OrderTag orderTag) {
+		orderTag.dose = orderTag.drug.getMinimumDailyDose();
+	}
+}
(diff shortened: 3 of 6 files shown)


api-1.10/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement1_10.java
```diff
@@ -417,7 +417,7 @@ protected void enterOrder(FormEntrySession session, OrderTag oldOrderTag) {
 		drugOrder.setCareSetting(Context.getOrderService().getCareSetting(orderTag.careSettingId));
 		OrderFrequency orderFrequency = Context.getOrderService().getOrderFrequency(Integer.valueOf(orderTag.frequency));
 		drugOrder.setFrequency(orderFrequency);
-		drugOrder.setDateActivated(orderTag.startDate);
+		drugOrder.setDateActivated(session.getEncounter().getEncounterDatetime());
```
mks-d, line 420: @mogoodrich beyond 1.10.x this does not work in ENTER mode because the drug order's date activated must be:
1. On or after the encounter datetime.
2. Before `Date.now()` (not a future datetime.)

... the only possible option in ENTER mode is to set it to be exactly the encounter date time. This is actually what's done in OE OWA.

api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
```diff
@@ -0,0 +1,23 @@
...
+	@Override
+	protected Date getDrugOrderStartDate(DrugOrder dod) {
+		return dod.getDateActivated();
+	}
+	
+	@Override
+	protected void setOrderTagDose(OrderTag orderTag) {
+		orderTag.dose = orderTag.drug.getMinimumDailyDose();
+	}
+}
```
mks-d, line 23: Ideally in ENTER mode we want the submission element to remove the start date widget since it cannot be used.

api-1.10/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement1_10.java
mogoodrich: Technically if you were retrospectively entered an encounter in ENTER mode you could set the order date differently than the encounter date as long as it was after the encounter date and before now, correct?

  (That being said, I'm fine with just making it be the encounter date, as that is probably the 98% use case)

api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
mogoodrich: Yeah, if I'm understanding everything that makes sense to me... I haven't seen a screen shot of this widget in some time.

Also, you can't edit a date activated as far as I know, so what would you be doing in EDIT mode?  You could add a new order to the encounter, I guess, but wouldn't that still fall under the same restrictions as ENTER mode?

api/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement.java
```diff
@@ -373,7 +373,7 @@ protected void populateDrugOrderValuesFromDB(FormEntryContext context, Boolean u
 	    				    if (((CheckboxWidget) drugWidget).getValue().equals(drugOrder.getDrug().getDrugId().toString()))
 	    				        ((CheckboxWidget) drugWidget).setInitialValue("CHECKED");
 	    				}
-	    				startDateWidget.setInitialValue(drugOrder.getStartDate());
+	    				startDateWidget.setInitialValue(getDrugOrderStartDate(drugOrder));
```
mogoodrich, line 376: So you are allowed to edit the start date of existing order?  Does this work?  (Seems like you are just modifying existing code, so I guess it does?)

api-1.10/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement1_10.java
mseaton: @Ruhanga / @mks-d - I agree that we should always set dateActivated to the encounterDatetime in ENTER mode.  And I think this should be true whether entering a retrospective encounter or not (eg. the order should always be activated in that encounter).  

The one use case where I think we would still want to support accepting a custom orderTag.startDate would be if it is after the encounterDatetime, in which case it should serve to set the "scheduledDate" property on the drug order, and should cause us to set the order urgency to "ON_SCHEDULED_DATE".

This allows us to retain support for "future orders", which is lost here.

api-2.0/pom.xml
```diff
@@ -23,13 +23,23 @@
...
             <groupId>org.openmrs.api</groupId>
             <artifactId>openmrs-api</artifactId>
             <version>${openMRSVersion}</version>
             <scope>provided</scope>
         </dependency>
+        
+        <dependency>
+			<groupId>org.openmrs.web</groupId>
+			<artifactId>openmrs-web</artifactId>
+		</dependency>
```
mseaton, line 42: This needs to be removed.  Nothing in an api jar can contain a dependency on a web artifact.

api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class DrugOrderSubmissionElement2_0 extends DrugOrderSubmissionElement1_10 {	
+	
+	public DrugOrderSubmissionElement2_0(FormEntryContext context, Map<String, String> parameters) {
+		super(context, parameters);
+	}
+	
+	@Override
+	protected Date getDrugOrderStartDate(DrugOrder dod) {
+		return dod.getDateActivated();
```
mseaton, line 16: We should use "dod.getEffectiveStartDate() here, as this will return the scheduledDate if set, otherwise the dateActivated.

api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
```diff
@@ -0,0 +1,23 @@
...
+	}
+	
+	@Override
+	protected void setOrderTagDose(OrderTag orderTag) {
+		orderTag.dose = orderTag.drug.getMinimumDailyDose();
```
mseaton, line 21: I think for 2.x, we just leave this as an empty implementation and don't do anything here.  It is not apparent to me that "minimumDailyDose" is the what we should be using as a default value for the dose of a given drug.

api-2.0/src/main/java/org/openmrs/module/htmlformentry/element/DrugOrderSubmissionElement2_0.java
mseaton: Per my comment above, I think we'd keep the startDateWidget, but maybe default it to the encounterDatetime and only allow it to be changed to something on or after that date.

For EDIT mode, I don't think the UI necessarily has to change, but we will need to properly handle this as a REVISE or DISCONTINUE/NEW order.

mseaton (changes requested):
@Ruhanga and @mks-d - sorry for the late review on this.  I need to add this support in myself and was reminded of this PR.  Not sure what your current cycles are for working on this.  How do you feel about me merging this into a shared branch and collaborating with you on it?

mseaton:
This PR is no longer valid given the extensive refactoring in progress with the drugOrder tag

PR #179, opened 2020-05-12
HTML-731: Fixed htmlFormEntry.js numbers validation.

shubhamsc:
https://issues.openmrs.org/browse/HTML-731


shubhamsc:
@mogoodrich @mks-d Please Review

rrameshbtech:
>I believe "val" is a string, so I fear Number.isInteger might always return false based on this line in the function docs:
> Number.isInteger('10'); // false

Yes @mogoodrich. `val` is string. So we are boxing with Number `val = Number(val);` before we do validation.

mogoodrich:
Ah, I missed that line thanks @rrameshbtech

rrameshbtech:
Thanks @mogoodrich. We have tested manually & ensured it is working as expected.

PR #180, opened 2020-05-14
HTML-725: <condition/> to support VIEW and EDIT modes.

samuelmale:
Ticket: https://issues.openmrs.org/browse/HTML-725

### DIFF
--- a/api-2.3/pom.xml
+++ b/api-2.3/pom.xml
@@ -7,13 +7,13 @@
         <version>3.11.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>htmlformentry-api-2.2</artifactId>
+    <artifactId>htmlformentry-api-2.3</artifactId>
     <packaging>jar</packaging>
-    <name>htmlformentry Module api 2.2</name>
-    <description>2.2 api project for htmlformentry</description>
+    <name>htmlformentry Module api 2.3</name>
+    <description>2.3 api project for htmlformentry</description>
 
     <properties>
-        <openMRSVersion>2.2.0</openMRSVersion>
+        <openMRSVersion>2.3.1-SNAPSHOT</openMRSVersion>
     </properties>
 
     <dependencies>
@@ -21,8 +21,21 @@
             <groupId>${project.parent.groupId}</groupId>
             <artifactId>${project.parent.artifactId}-api</artifactId>
             <version>${project.parent.version}</version>
+            <classifier>tests</classifier>
+			<exclusions>
+				<exclusion>
+					<artifactId>javassist</artifactId>
+					<groupId>javassist</groupId>
+				</exclusion>
+			</exclusions>
         </dependency>
 
+		<dependency>
+			<groupId>org.openmrs.module</groupId>
+			<artifactId>htmlformentry-api-tests</artifactId>
+			<version>${project.parent.version}</version>
+		</dependency>
+		
         <dependency>
             <groupId>org.openmrs.api</groupId>
             <artifactId>openmrs-api</artifactId>
@@ -30,6 +43,19 @@
             <scope>provided</scope>
         </dependency>
        
+       <dependency>
+			<groupId>org.openmrs.api</groupId>
+			<artifactId>openmrs-api</artifactId>
+			<version>${openMRSVersion}</version>
+			<classifier>tests</classifier>
+			<exclusions>
+				<exclusion>
+					<artifactId>javassist</artifactId>
+					<groupId>javassist</groupId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+		
 		<dependency>
 			<groupId>org.openmrs.test</groupId>
 			<artifactId>openmrs-test</artifactId>
@@ -40,6 +66,37 @@
         <dependency>
 			<groupId>org.openmrs.web</groupId>
 			<artifactId>openmrs-web</artifactId>
+			<version>${openMRSVersion}</version>	
+		</dependency>
+		
+		<!-- jars required only to run test scripts (scope=test) -->
+		<dependency>
+			<groupId>org.codehaus.groovy</groupId>
+			<artifactId>groovy</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>cglib</groupId>
+			<artifactId>cglib</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>xalan</groupId>
+			<artifactId>xalan</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>joda-time</groupId>
+			<artifactId>joda-time</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.joda</groupId>
+			<artifactId>joda-convert</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.codehaus.jackson</groupId>
+			<artifactId>jackson-mapper-asl</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.codehaus.jackson</groupId>
+			<artifactId>jackson-core-asl</artifactId>
 		</dependency>
 		
 		<dependency>
@@ -52,6 +109,18 @@
     </dependencies>
 
     <build>
+	  <plugins>
+	      <plugin>
+	        <groupId>org.apache.maven.plugins</groupId>
+	        <artifactId>maven-compiler-plugin</artifactId>
+	        <version>3.8.1</version>
+	        <configuration>
+	          <source>1.8</source>
+	          <target>1.8</target>
+	        </configuration>
+	      </plugin>
+	    </plugins>
+	    
         <resources>
             <resource>
                 <directory>src/main/resources</directory>
--- a/api-2.3/pom.xml
+++ b/api-2.3/pom.xml
@@ -9,7 +9,7 @@
 
     <artifactId>htmlformentry-api-2.3</artifactId>
     <packaging>jar</packaging>
-    <name>htmlformentry Module api 2.3</name>
+    <name>HTML Form Entry API 2.3</name>
     <description>2.3 api project for htmlformentry</description>
 
     <properties>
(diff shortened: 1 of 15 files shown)


api-2.3/pom.xml
```diff
@@ -7,29 +7,55 @@
         <version>3.11.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>htmlformentry-api-2.2</artifactId>
+    <artifactId>htmlformentry-api-2.3</artifactId>
     <packaging>jar</packaging>
-    <name>htmlformentry Module api 2.2</name>
-    <description>2.2 api project for htmlformentry</description>
+    <name>htmlformentry Module api 2.3</name>
```
mks-d, line 12: Whilst you're at it, let's rename other modules to achieve normalization:
```shell
[INFO] HTML Form Entry ....................................
[INFO] HTML Form Entry API ................................
[INFO] htmlformentry Module api 1.9 .......................
[INFO] HTML Form Entry API 1.10 ...........................
[INFO] htmlformentry Module api 2.0 .......................
[INFO] HTML Form Entry API Tests ..........................
[INFO] htmlformentry Module api 2.3 .......................
[INFO] HTML Form Entry OMOD ...............................
```
So:
* `htmlformentry Module api 1.9` → `HTML Form Entry API 1.9`
* ... etc

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,204 @@
...
+	
+	private MessageSourceService mockedMessageSourceService() {
+		MessageSourceService ret = mock(MessageSourceService.class);
+		when(ret.getMessage("coreapps.conditionui.condition")).thenReturn("Condition");
+		when(ret.getMessage("coreapps.conditionui.active.label")).thenReturn("active");
+		when(ret.getMessage("coreapps.conditionui.inactive.label")).thenReturn("inactive");
+		when(ret.getMessage("htmlformentry.conditionui.historyOf.label")).thenReturn("history-of");
+		when(ret.getMessage("coreapps.conditionui.status")).thenReturn("Status");
+		when(ret.getMessage("coreapps.conditionui.onsetdate")).thenReturn("Onset Date");
+		when(ret.getMessage("coreapps.stopDate.label")).thenReturn("End Date");
```
mks-d, line 192: I understand that this is just test-only, but let's make sure that all i18n messages are contained in the HFE module itself. Because in fact it should be able to work without Core Apps being provided.

samuelmale:
> Sorry, I'm super familar with the Condition functionality.. what is the difference between the 2.2 Condition and the 2.3 Condition that means we need to have two different tag handlers?

@mogoodrich Condition tag handler for 2.2 has an incorrect support for `<condition/>`. The tag didn't have support for modes like edit and view. Since this was a relatively a new feature to HFE, we felt it safe and okay to have a new one that gracefully supports conditions. So the short answer is, no one wants to use the one in 2.2 since it's less featured and incomplete.

mks-d:
@samuelmale you forgot to say that it's only from Core 2.3 that `condition` is linked to `encounter`.

@mogoodrich this means that a full HFE support for managing conditions within HFE can only be achieved under 2.3 compatibility.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -0,0 +1,453 @@
...
+	}
+
+	@Override
+	public String generateHtml(FormEntryContext context) {
+		if (mss == null) {
+			mss = Context.getMessageSourceService();
+		}
+		initializeExistingCondition(context);
+		StringBuilder ret = new StringBuilder();
+		ret.append("<div id=\"htmlformentry-condition\">");
```
mks-d, line 114: In regards to this `id`, what's going to happen when two condition tags are used in the same form?

samuelmale: Nice catch, we should be smart enough here. Gonna fix this

mogoodrich:
Ah, thanks @samuelmale @mks-d ... I was going to suggest we might want to consider replacing the 2.2 implementation entirely, but it sounds like we aren't doing that because we need the functionality to tie to an encounter?

mks-d:
>I was going to suggest we might want to consider replacing the 2.2 implementation entirely

@mogoodrich the plan is to have only a 2.3 submodule, no more a 2.2 submodule. In other words the `<condition/>` tag will only be supported from Core 2.3.1. As for what was released within HFE 3.10.0, we may need to patch it as new 3.10.1 release...

mogoodrich:
@mks-d  fine with me!

samuelmale:
@mks-d I addressed the comments above.

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.openmrs.Patient;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.htmlformentry.RegressionTestHelper;
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+public class ConditionTagTest extends BaseModuleContextSensitiveTest {
+	
+	// field names
+	private String searchWidgetForCurrectCondition = "w7";
```
mks-d, line 25: → `searchWidgetIdForCurrentCondition `

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+	
+	private String statusWidgetForCurrectCondition = "w9";
```
mks-d, line 27: → `statusWidgetIdForCurrentCondition`

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+	
+	private String onsetDateWidgetForCurrectCondition = "w11";
```
mks-d, line 29: → `onsetDateWidgetIdForCurrentCondition`

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+			}
+			
+			@Override
+			public String[] widgetLabels() {
+				return new String[] { "Date:", "Location:", "Provider:" };
+			}
+			
+			@Override
+			public void testBlankFormHtml(String html) {
+				System.out.println(html);
```
mks-d, line 60: This should go away.

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+			public boolean doEditEncounter() {
+				return true;
+			}
+			
+			@Override
+			public void setupEditRequest(MockHttpServletRequest request, Map<String, String> widgets) {
+				// do edit the current condition
+				request.setParameter(searchWidgetForCurrectCondition, "Epilepsy");
+				request.setParameter(searchWidgetForCurrectCondition + "_hid", "3476");
+				request.setParameter(statusWidgetForCurrectCondition, "active");
```
mks-d, line 116: Those 3 lines should go.

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java
```diff
@@ -0,0 +1,193 @@
...
+			
+			@Override
+			public void testEditedResults(SubmissionResults results) {
+				// setup
+				Set<Condition> conditions = results.getEncounterCreated().getConditions();
+				
+				results.assertNoErrors();
+				Assert.assertEquals(2, conditions.size());
+				
+				Condition currentCondition = results.getEncounterCreated().getConditions().iterator().next();
```
mks-d, line 129: Let's avoid this iterator, let's fetch the first or second element from the array as done [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/180/files#diff-79c0798270724e704144ef7cc75b8eddR92-R103).

mogoodrich:
Will take a look @mks-d ... we are on critical path for a big release this weekend, so may not get to it until next week.

api-2.3/pom.xml
```diff
@@ -7,29 +7,55 @@
...
+    <artifactId>htmlformentry-api-2.3</artifactId>
     <packaging>jar</packaging>
-    <name>htmlformentry Module api 2.2</name>
-    <description>2.2 api project for htmlformentry</description>
+    <name>HTML Form Entry API 2.3</name>
+    <description>2.3 api project for htmlformentry</description>
 
     <properties>
-        <openMRSVersion>2.2.0</openMRSVersion>
+        <openMRSVersion>2.3.1-SNAPSHOT</openMRSVersion>
```
mogoodrich, line 16: We can update to 2.3.1 now since I did the release last week.

api-2.3/pom.xml
```diff
@@ -7,29 +7,55 @@
...
             <groupId>${project.parent.groupId}</groupId>
             <artifactId>${project.parent.artifactId}-api</artifactId>
             <version>${project.parent.version}</version>
+            <classifier>tests</classifier>
+			<exclusions>
+				<exclusion>
+					<artifactId>javassist</artifactId>
+					<groupId>javassist</groupId>
+				</exclusion>
+			</exclusions>
```
mogoodrich, line 30: What does this exclusion do?

samuelmale: To be sure we don't have conflicting `javassist` jars in the classpath.

samuelmale:
@mogoodrich I have upgraded to `2.3.1`.

PR #181, opened 2020-05-15
Configuring formatter + formatting .java files.

mks-d:
@mogoodrich sorry, there's no ticket for this.

Are you ok to let this go through?


mogoodrich:
I'd never set up the OpenMRS formatter in the past because I feel like there's never been good time and there were some formatting things it did that I wasn't crazy about.

I'm fine with taking a stab at it now... however, it *will* make any existing pull requests either a nightmare to review or merge, right?  I think that may have been my reasoning for never setting this up before.

ibacher:
> I'd never set up the OpenMRS formatter in the past because I feel like there's never been good time and there were some formatting things it did that I wasn't crazy about.

I [recently added support](https://github.com/openmrs/openmrs-core/commit/4be9c5e30f92a54d99cc4d970101e367ceeafffa) for disabling the code formatter.

> however, it will make any existing pull requests either a nightmare to review or merge, right?

More likely it will just give you a bunch of merge conflicts.

mks-d:
>I'd never set up the OpenMRS formatter in the past because I feel like there's never been good time and there were some formatting things it did that I wasn't crazy about.

@mogoodrich agreed, it depends which file types you let it touch, and which version that is being used. Its formatting when using old versions of Core might be annoying at times indeed, see [this upgrade](https://github.com/mekomsolutions/openmrs-module-initializer/commit/83a8974ecc23b1c1bdf1fb33e982a177324e1e8f#diff-e701cf5b0c3f7ab6625a2e915d5adf0b) for instance.

mogoodrich:
So @ibacher are you suggesting we don't use the formatter in this case?  I'm been resistance to applying to HTML Form Entry, but if we still think it's best practice I'm okay with this going through.

ibacher:
@mogoodrich I'd actually recommend using the formatter. It's far from perfect, but (personally) I find the consistent code style ultimately makes code reviews easier.

I was just pointing out that there is a work-around when it does something truly awful and that while it won't necessarily make existing PRs any more difficult to review, but it will likely require either the PR author or the merger to do some manual reconciliation.

mogoodrich:
Okay, fine with me then @ibacher  @mks-d ... though @mks-d  we might want to time it so that we merge it in when the majority of the recent PRs have been merged in, to save people some work (I think the majority are things your team are working on... I don't have any in particular I'm waiting on)?

mogoodrich:
Of course, this build fails now on PIH Bamboo because we are using an old version of Maven and it looks like the formatter plugin you added needs at least Maven 3.3.9.

We should really upgrade the version of Maven on our Bamboo server, but @mks-d  are you using a different plugin or plugin version than is used in Core?

mogoodrich:
@mks-d  never mind... we've got a later version of Maven installed on our servers, I just didn't realzie that that particular job wasn't configured to us it.

mks-d:
@mogoodrich yes I tend to use [revelc/formatter-maven-plugin](https://github.com/revelc/formatter-maven-plugin). Ok I see that OpenMRS' Bamboo is also fine, Maven was recently upgraded by @cintiadr 👍

PR #192, opened 2020-07-22
HTML-742: Condition tag to support formPath attribute.

icrc-loliveira:
## Issue I worked on

see https://issues.openmrs.org/browse/HTML-742


tendomart:
@icrc-loliveira  Travis is having some issues
Line 3271 at [ERROR] /home/travis/build/openmrs/openmrs-module-htmlformentry/api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java:[92,34] cannot find symbol

  symbol:   method setFormField(java.lang.String,java.lang.String)


Did you run " mvn clean package " before doing the commit. It would save you time.

mks-d:
@tendomart this is draft PR to start sharing code, no formal review is expected as of yet.

tendomart:
Ooh Thanks @mks-d  for letting me know.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -87,6 +89,7 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
 			}
 			condition.setPatient(session.getPatient());
 			condition.setAdditionalDetail(formFieldName);
+			condition.setFormField("htmlformentry", formFieldPath);
```
mks-d, line 92: `"htmlformentry"` → `"HtmlFormEntry"` as per [here](https://talk.openmrs.org/t/hfe-conditions-should-have-a-pointer-for-the-form-field-it-was-created-through/29110/54?u=mksd), also make it a constant `FORM_NAMESPACE` in [`HtmlFormEntryConstants`](https://github.com/openmrs/openmrs-module-htmlformentry/blob/77e7f343bd7361762676d733031af220a478e1d2/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryConstants.java#L5-L40).

api-2.3/src/main/java/org/openmrs/module/htmlformentry/handler/ConditionTagHandlerSupport2_3.java
```diff
@@ -8,29 +8,30 @@
...
+
+    @Override
+    public String getSubstitution(FormEntrySession session, FormSubmissionController controller,
+                                  Map<String, String> attributes) {
+        conditionElement = new ConditionElement();
+        String required = attributes.get("required");
+        if (required != null) {
+            conditionElement.setRequired(required.equalsIgnoreCase("true"));
+        }
+        String formFieldName = attributes.get("formFieldName");
```
mks-d, line 27: The attribute `"formFieldName"` needs to become `"formPath"`. This also needs to be changed [here](https://github.com/openmrs/openmrs-module-htmlformentry/blob/77e7f343bd7361762676d733031af220a478e1d2/api/src/test/resources/org/openmrs/module/htmlformentry/include/conditionForm.xml#L7-L8). And then [`ConditionTagTest`](https://github.com/openmrs/openmrs-module-htmlformentry/blob/77e7f343bd7361762676d733031af220a478e1d2/api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionTagTest.java#L20) will have to assert that the saved conditions do carry the form namespace and path as expected.

icrc-loliveira: So,  the goal is to replace the `formFieldName` tag with the `formPath`, insted of just add it? 
Will this not break the compatibility with older versions?

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
icrc-loliveira: Done :)

api-2.3/src/test/java/org/openmrs/htmlformentry/element/ConditionElementTest.java
```diff
@@ -234,7 +234,7 @@ public void handleSubmission_shouldSupportFormField() {
 		Condition condition = conditions.iterator().next();
 		Assert.assertEquals(ConditionClinicalStatus.ACTIVE, condition.getClinicalStatus());
 		Assert.assertThat(condition.getCondition().getCoded().getId(), is(1519));
-		Assert.assertEquals("htmlformentry^fieldName", condition.getFormNamespaceAndPath());
+		Assert.assertEquals("HtmlFormEntry^fieldName", condition.getFormNamespaceAndPath());
```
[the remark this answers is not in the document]
icrc-loliveira: Could you please validate that I made the change as you expected?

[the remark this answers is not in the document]
icrc-loliveira: Done.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -163,15 +165,15 @@ private Condition bootstrap(FormEntryContext context) {
...
 	 */
 	private void initializeExistingCondition(FormEntryContext context) {
 		if (context.getMode() != Mode.ENTER) {
 			Set<Condition> conditions = context.getExistingEncounter().getConditions();
 			for (Condition candidate : conditions) {
-				if (candidate.getAdditionalDetail().equals(formFieldName)) {
+				if (candidate.getFormFieldPath().equals(formFieldPath)) {
 					this.existingCondition = candidate;
 					return;
 				}
```
icrc-loliveira: Hi @mks-d,
Here I am having some troubles.
The `candidate.getFormFieldPath()`  will return what we expected (something like `HtmlFormEntry^MyForm.1.0/my_condition_tag-0`), but I we only have access to the `formFieldPath` (something like `my_condition_tag`) and I cannot calculate the full path again because here we don't have access to the `FormEntrySession` to get the Form info.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -163,15 +165,15 @@ private Condition bootstrap(FormEntryContext context) {
...
 	 */
 	private void initializeExistingCondition(FormEntryContext context) {
 		if (context.getMode() != Mode.ENTER) {
 			Set<Condition> conditions = context.getExistingEncounter().getConditions();
 			for (Condition candidate : conditions) {
-				if (candidate.getAdditionalDetail().equals(formFieldName)) {
+				if (candidate.getFormFieldPath().equals(formPath)) {
 					this.existingCondition = candidate;
 					return;
 				}
```
icrc-loliveira, line 179: Hi @mks-d,

Here I am having some troubles.
The `candidate.getFormFieldPath()`  will return what we expected (something like `HtmlFormEntry^MyForm.1.0/my_condition_tag-0`), but I we only have access to the `formFieldPath` (something like `my_condition_tag`) and I cannot calculate the full path again because here we don't have access to the `FormEntrySession` to get the Form info.

mks-d: 1. The widget on the form looks something like `<condition formPath="my_condition_tag"/>`
1. The `Condition`'s candidate instance's form namespace and path is something like `HtmlFormEntry^MyForm.1.0/my_condition_tag-0`.

You need to extract the [control id](https://talk.openmrs.org/t/hfe-conditions-should-have-a-pointer-for-the-form-field-it-was-created-through/29110/24?u=mksd) part out of the form namespace and path, and check whether it matches the form path attribute of the widget. In the end it will be something like this I believe:
```java
if (getControlId(candidate.getFormFieldPath()).equals(formPath)) {
```

icrc-loliveira: Done.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -86,7 +88,7 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
 				condition.setEndDate(endDateWidget.getValue(context, submission));
 			}
 			condition.setPatient(session.getPatient());
-			condition.setAdditionalDetail(formFieldName);
+			condition.setFormField(FORM_NAMESPACE, session.generateFormField(formPath, "0"));
```
icrc-loliveira, line 91: Hi @mks-d, 
This is the last part that I don' know how it should work.
As you can see I have the count hard-coded at `0`. How should this counter work?

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -457,4 +459,20 @@ public void setExistingCondition(Condition existingCondition) {
...
+	 * @return
+	 */
+	public String getControlId(Condition condition) {
+		if (condition.getFormFieldPath() == null) {
+			return null;
+		} else {
+			String controlId = condition.getFormFieldPath().split("/")[1];
+			return controlId.split("-")[0];
+		}
+	}
```
icrc-loliveira, line 477: Hi @mks-d,

Unfortunately  the `HtmlFormEntryUtil` is in the `htmlFormEntry-api` that doesn't have access to the `Condition` class. Not only that, but it also need to have access to the `condition.getFormFieldPath()` that have only been added in the last version. so this method will have to be somewhere in the  `htmlFormEntry-api-2.3`.

Let me know how I should proceed.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -1185,4 +1185,28 @@ public String getEncounterEncounterTypeName() {
...
+	
+	/**
+	 * Create form field based on the form name, form version, form field path and a counter. The form
+	 * field will have the following format: <MyForm.1.0/my_condition_tag-0>
+	 *
+	 * @param formPath
+	 * @param count
+	 * @return
+	 */
+	public String generateFormField(String formPath, Integer count) {
```
mks-d, line 1197: The inputs are:
* control id
* control counter

And this function returns the form path or even rather the control path _in_ the form. So please change it to
```java
public String generateControlFormPath(String controlId, Integer controlCounter) {
```

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -1185,4 +1185,28 @@ public String getEncounterEncounterTypeName() {
...
+		String formField = "";
+		
+		// Create form path
+		if (this.getForm() != null) {
```
mks-d, line 1201: Do the contrary, throw early if the form is null, something like:
```java
if (this.getForm() == null) {
  throw new IllegalStateException("The form entry session has a null form.");
}
```

icrc-loliveira: Done

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
icrc-loliveira: Done.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -457,4 +459,20 @@ public void setExistingCondition(Condition existingCondition) {
...
 	}
+
+	/**
+	 * Return control id from Condition.formFieldPath.
+	 * HtmlFormEntry^MyForm.1.0/<b>my_condition_tag</b>-0
+	 *
+	 * @param condition
+	 * @return
+	 */
+	public String getControlId(Condition condition) {
```
mks-d, line 470: As far as I'm concerned you could create a `HtmlFormEntryUtil2_3` in the **api-2.3** submodule that provides 2.3-specific static helpers. However the signature should be:
```java
public String getControlId(FormRecordable openmrsData) {
```

icrc-loliveira: Done

api-2.3/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil2_3.java
```diff
@@ -0,0 +1,26 @@
...
+	 * @return The control id
+	 */
+	public static String getControlId(FormRecordable openmrsData) {
+		if (openmrsData.getFormFieldPath() == null) {
+			return null;
+		} else {
+			String controlId = openmrsData.getFormFieldPath().split("/")[1];
+			return controlId.split("-")[0];
+		}
+	}
```
mks-d, line 24: That will definitely not work when the form field path is something like `"HtmlFormEntry^MyForm.1.0/my-condition-tag-0"`...
We need a new test class and test cases for this method.

As for the method's logic itself, it will have to look for the last hyphen and ensure that the RHS can be parsed into an `Integer`... etc etc.

You don't need the `else` statement by the way, let's remove it.

We also need to think carefully about what to do if `openmrsData.getFormFieldPath() == null`, I would probably throw an [`IllegalStateException`](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html), with a message like
>`"A form recordable object was found to have no form namespace and path set, its control id in the form could not be determined."`

Furthermore change the `if` test to:
```java
if (StringUtils.isEmpty(openmrsData.getFormFieldPath())) {
```

icrc-loliveira: Nice catch :)
It should be Ok now.

icrc-loliveira: Nice catch :)
It should be OK now.

api-2.3/src/test/java/org/openmrs/htmlformentry/HtmlFormEntryUtil2_3Test.java
```diff
@@ -0,0 +1,67 @@
...
+	@Test(expected = IllegalStateException.class)
+	public void getControlId_shouldThrowIllegalStateException() {
+		
+		// Prepare parameters
+		Obs observation = new Obs();
+		
+		// Test
+		String controlId = HtmlFormEntryUtil2_3.getControlId(observation);
+		
+	}
```
mks-d, line 66: Could  you add test cases for a form paths like those two:
1. `"MyForm.1.0-42fcd95f/my_condition_tag-0"` 
1. `"MyForm.1.0/my_condition_tag-123-0"`

icrc-loliveira: Done :)

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
```diff
@@ -35,7 +37,7 @@
 	
 	private boolean required;
 	
-	private String formFieldName;
+	private String controlId;
```
mks-d, line 40: @mogoodrich ideally all the logic around the newly introduced control id should be in a shared base class that `ConditionElement` extends.
Do you think it is ok to **not** do that this time since this is the first and only implementation of an element dealing with its control id, and to introduce the new shared based class whenever a second implementation is introduced?

api-tests/src/test/java/org/openmrs/module/htmlformentry/FormEntrySessionTest.java
```diff
@@ -92,4 +93,24 @@ public void createForm_shouldReturnCorrectXmlWithACompoundExpressionInAnExcludeI
 		FormEntrySession session = new FormEntrySession(patient, htmlform, null);
 		Assert.assertEquals("<div class=\"htmlform\"></div>", session.getHtmlToDisplay());
 	}
+	
+	/**
+	 * @see {@link FormEntrySession#generateFormField(String, String)}
+	 */
+	@Test
+	@Verifies(value = "should return the form field with thwe form name, form version, form path and a counter", method = "generateFormField(String, String)")
+	public void createForm_shouldReturnTheFormFiled() throws Exception {
```
mks-d, line 102: There is a typo: `createForm_shouldReturnTheFormFiled`.
Anyway, please rename this to "form path" (not form field).

api-tests/src/test/java/org/openmrs/module/htmlformentry/FormEntrySessionTest.java
```diff
@@ -92,4 +93,24 @@ public void createForm_shouldReturnCorrectXmlWithACompoundExpressionInAnExcludeI
...
+		String excludeText = "Patient age is valid";
+		String htmlform = "<htmlform><excludeIf velocityTest=\"$patient.age >= 1 && $patient.age <= 120 \">" + excludeText
+		        + "</excludeIf></htmlform>";
+		FormEntrySession session = new FormEntrySession(patient, htmlform, null);
+		Form form = new Form();
+		form.setName("MyForm");
+		form.setVersion("1.0");
+		session.setForm(form);
+		
+		String formField = session.generateControlFormPath("my_condition_tag", 0);
```
mks-d, line 112: `formPath`

api-2.3/pom.xml
```diff
@@ -13,7 +13,7 @@
     <description>2.3 api project for htmlformentry</description>
 
     <properties>
-        <openMRSVersion>2.3.1</openMRSVersion>
+        <openMRSVersion>2.3.2-SNAPSHOT</openMRSVersion>
```
mogoodrich, line 16: For what it's worth, this means we won't be able to release HFE until we do the 2.3.2 Core release... are you guys planning on doing that at some point?

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
mogoodrich: Sure, though we are likely going to be adding this to Obs in the near future, right?

api-2.3/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil2_3.java
```diff
@@ -0,0 +1,39 @@
...
+	 * @param openmrsData The form recordable OpenMRS data object
+	 * @return The control id
+	 */
+	public static String getControlId(FormRecordable openmrsData) {
+		
+		// Validate if getFormFieldPath exists.
+		if (StringUtils.isEmpty(openmrsData.getFormFieldPath())) {
+			throw new IllegalStateException(
+			        "A form recordable object was found to have no form namespace and path set, its control id in the form could not be determined.");
+		}
```
mogoodrich, line 25: for what it's worth, when we expand Obs to support a form path, we will need change this, because we need to support legacy forms/obs that don't have a control Id.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/ConditionElement.java
mks-d: @mogoodrich correct but that's out of our in-house scope, at least for now.

api-2.3/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil2_3.java
mks-d: @icrc-loliveira I suggest that we return `null` early then, and let the caller throw exceptions. In our case the `IllegalStateException` will this be thrown in `ConditionElement`.

Please adjust the JavaDoc accordingly and add a test case in `ConditionElementTest`.

api-tests/src/test/java/org/openmrs/module/htmlformentry/FormEntrySessionTest.java
mks-d: @icrc-loliveira ↑

api-2.3/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil2_3.java
```diff
@@ -0,0 +1,38 @@
...
+ * HTML Form Entry utility methods with Core 2.3 compatibility
+ */
+public class HtmlFormEntryUtil2_3 {
+	
+	/**
+	 * Returns the control id part out of an OpenMRS data object's form namespace and path. Eg:
+	 * "my_condition_tag" out of "HtmlFormEntry^MyForm.1.0/my_condition_tag-0"
+	 *
+	 * @param openmrsData The form recordable OpenMRS data object
+	 * @return The control id
```
mks-d, line 17: >`The control id or null if the form recordable has no form namespace and path set`

icrc-loliveira: Done.

PR #202, opened 2020-09-18
HTML-750: Improve "when" functionality to support radio buttons

mogoodrich:


omod/src/main/webapp/resources/htmlForm.js
```diff
@@ -11,6 +11,12 @@
           val = $("#"+$(this).attr("id")+"_hid").val();
         }
 
+        // handle differently radio sets since the obs value is associated with the name, not id
+        // (in radio sets, each radio button has an id in format "w123_x" but all have the same name in format "w123")
+        if ($(this).attr('id') && $(this).attr('id').indexOf("_") !== -1) {
```
brandones, line 16: Are there other conditions under which an obs element might wind up with an id with an underscore in it?

brandones:
Are there tests for this kind of thing?

mogoodrich:
Good question... test coverage for the core Java functionality in HFE is actually (relatively) comprehensive but non-existent for the JS stuff.

omod/src/main/webapp/resources/htmlForm.js
mogoodrich: I hope not?  :)

No, good question... I'm not sure, though I would hope that they follow the same pattern.

That being said, this block of code is pretty targeted... it's in function that handles the "when" control in HFE, and there's already a special case for autocomplete.  I think if there *is* another element that uses the underscore, that element would not have worked with the "when" control anyway, and we can address it if and when we have to use that functionality.

brandones (approved):

PR #203, opened 2020-09-29
HTML-751 - Increase minimum version to 2.1.0 and enable testing support of all 2.x versions

mseaton:
This is a commit to start a new major release line that increases the minimum required OpenMRS version to 2.1.0.  The driving reason for this is to ensure full and proper test coverage for the latest releases of OpenMRS, which proved difficult while also supporting all releases back to 1.9, and to improve future maintainability of the module.  This also has the following backwards-incompatible changes:
* Removes support for logic expressions (velocity only supported)
* Removes support for the "standardRegimen" element / tag with Drug Orders, which was removed in OpenMRS 2.x
There are significant, substantive changes to the project structure (removing several Maven submodules) and to the unit tests in order to maintain only a single version of the test data file where possible, and many "compatibility" classes were able to be removed due to no longer needing to support features prior to 2.1.
OpenMRS 2.1 was chosen over OpenMRS 2.0 due to a bug in how Observations are saved in 2.0 that prevents the module from working successfully.


api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
```diff
@@ -582,18 +578,12 @@ public void testResults(SubmissionResults results) {
...
-				PatientProgram patientProgram = getPatientProgramByState(results.getPatient(), state, PAST_DATE);
-				PatientState patientState = getPatientState(patientProgram, state, PAST_DATE);
+				PatientProgram patientProgram = getPatientProgramByState(results.getPatient(), state, ONE_YEAR_AGO);
+				PatientState patientState = getPatientState(patientProgram, state, ONE_YEAR_AGO);
 				Assert.assertNotNull(patientProgram);
-				Assert.assertEquals(dateAsString(PAST_DATE), dateAsString(patientProgram.getDateEnrolled()));
-				Assert.assertEquals(dateAsString(PAST_DATE), dateAsString(patientState.getStartDate()));
-				Assert.assertEquals(dateAsString(DATE), dateAsString(patientState.getEndDate()));
-				
-				patientProgram = getPatientProgramByState(results.getPatient(), state, DATE);
```
mogoodrich, line 592: I know we discussed this, but why was this block removed again?  I would think this should still pass?

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
```diff
@@ -790,24 +783,13 @@ public boolean doViewEncounter() {
...
-			@Override
-			public boolean doEditEncounter() {
-				return true;
-			}
-			
-			@Override
-			public void testEditFormHtml(String html) {
-				Assert.assertTrue("Edit should contain current state: " + html,
-				    html.contains("selected=\"true\">MIDDLE STATE"));
-			}
```
mogoodrich, line 803: Another nit, but I think you were saying this fails because we don't have the submit for the edit working properly?

I don't think we need to fix this, but before you remove this can you just confirm that this assert passes properly and confirm that it's somewhere further downstream where this is failing?

mogoodrich (commented):
@mseaton a couple comments/questions on the workflow state test changes.

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
mseaton: Added this back in, using new test methods

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
mseaton: I've re-enabled this.  I just had to override a different method to prevent the edit submission from happening.

mseaton:
@mogoodrich - I've addressed the issues you raised in WorkflowStateTagTest.  I ended up writing some new test methods that simplify the code, so will be hard to see my fixes without a deep compare, but let me know if you have any further feedback prior to merging.

PR #204, opened 2020-09-29
HTML-745: Allow condition tag to be preset with a concept.

mks-d:
Forward-porting [HTML-745](https://issues.openmrs.org/browse/HTML-745) to the new 4.x branch.


mseaton (commented):
@mks-d - all looks good to me aside from the RegressionTest-data-openmrs-2.30.xml.  The new approach taken is not to maintain separate regression test data files per supported version.  Ideally you'd extend BaseHtmlFormEntryTest and use the new executeVersionedDataSet method to load in the existing regression test data like so:
`executeVersionedDataSet("org/openmrs/module/htmlformentry/data/RegressionTest-data-openmrs-2.1.xml");`
If you need additional test data from what is in here, ideally you'd just add it to this file.  And nothing should be added here that is already part of the standardTestDataset.xml that is from core.

api-2.3/src/test/java/org/openmrs/htmlformentry/ConditionTagTest.java
```diff
@@ -0,0 +1,262 @@
...
+import org.openmrs.module.htmlformentry.HtmlFormEntryUtil2_3;
+import org.openmrs.module.htmlformentry.RegressionTestHelper;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+public class ConditionTagTest extends BaseHtmlFormEntryTest {
+	
+	@Before
+	public void setup() throws Exception {
+		executeVersionedDataSet("org/openmrs/module/htmlformentry/data/RegressionTest-data-openmrs-2.3.xml");
+	}
```
mseaton, line 31: @mks-d - the goal of the "executeVersionedDataSet" method is so we only need to have one data set for all supported versions, not separate ones per version.  The reason the existing RegressionTest-data-openmrs-2.1.xml is named the way it is, is to let people know that it should be authored with 2.1 data model support.  Then, the implementation of executeVersionedDataSet is such that it will manipulate this single data set into the necessary compatibility for later OpenMRS versions.  Right now, this just means renaming the concept "precise" column to "allow_decimal" in OpenMRS 2.2+, but we'd add additional conversions as needed in order to only have to maintain one test data file.

So...you should be executing against the RegressionTest-data-openmrs-2.1.xml file here, not creating a new "2.3" version.

mseaton: See implementation here:
https://github.com/openmrs/openmrs-module-htmlformentry/blob/842502f02d0b3d85fa41792573db00a576230ffa/api/src/test/java/org/openmrs/module/htmlformentry/BaseHtmlFormEntryTest.java

api-2.3/src/main/java/org/openmrs/module/htmlformentry/element/ConditionElement.java
```diff
@@ -62,50 +69,60 @@
...
+		
+		CodedOrFreeText codedOrFreeText = new CodedOrFreeText();
+		if (presetConcept != null) {
+			codedOrFreeText.setCoded(presetConcept);
+		} else {
 			try {
-				int conceptId = Integer.parseInt((String) conditionSearchWidget.getValue(session.getContext(), submission));
-				conditionConcept.setCoded(Context.getConceptService().getConcept(conceptId));
-				
+				int conceptId = Integer.parseInt((String) conceptSearchWidget.getValue(session.getContext(), submission));
```
mseaton, line 79: Typically here you'd just use HtmlFormEntryUtil.getConcept(String id)

api-2.3/src/main/java/org/openmrs/module/htmlformentry/handler/ConditionTagHandler.java
```diff
@@ -14,15 +15,23 @@
...
+		
 		String controlId = attributes.get("controlId");
 		if (StringUtils.isBlank(controlId)) {
 			throw new IllegalArgumentException("Attribute controlId cannot be blank");
 		}
 		conditionElement.setControlId(controlId);
+		
+		String conceptId = attributes.get("conceptId");
+		if (!StringUtils.isEmpty(conceptId)) {
+			conditionElement.setPresetConcept(HtmlFormEntryUtil.getConcept(conceptId));
```
mseaton, line 32: This is what I was referring to above.  This is a bit cleaner.

mseaton (approved):
Looks good to me.  Just one minor comment suggestion, but not something that should hold up the PR.  We can merge in when ready.

PR #205, opened 2020-09-30
Set dateActivated and scheduledDate appropriately based on startDate …

mseaton:
…entered.  Ensure that a non-null OrderType is associated without relying on a specific uuid.


mseaton:
FYI - I'm merging this in prematurely as I need to get it up and running on a test server in my distribution, but please comment here with any post-commit changes needed.

PR #208, opened 2020-10-14
HTML-754: NumberFieldWidget does not throw validation exception if va…

mogoodrich:
…lue is has decimal component and isFloat=false

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/PatientDetailSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/PatientDetailSubmissionElement.java
@@ -13,17 +13,6 @@
  */
 package org.openmrs.module.htmlformentry.element;
 
-import javax.servlet.http.HttpServletRequest;
-import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
-
 import org.openmrs.Location;
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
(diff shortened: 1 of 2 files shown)


brandones (approved):

mseaton (commented):
FYI - I've made (in my in progress branch) some related changes to the NumberFieldWidget that might lead to a slightly different solution.  

What I think we should do is:

a) Change the return type of the getValue method to be a "Number".
b) Instead of parsing to a Double, parse to either an Integer or a Double depending on the value of floatingPoint true/false.
c) In the catch block for NumberFormatException, change the message based on whether floating point is true/false to say "Not a valid integer" or "Not a valid number" depending.

This will solve this issue, and also has tangential benefits that the returned value accurately reflects what was actually collected and intended by the widget.

mogoodrich:
@mseaton  that works for me... are you currently working on that as part of your commit, or should I work on this?

mseaton:
If you could just bang it in, that would be better.  My branch is interminable and would need to resolve conflicts here.  I'd rather just revert my change and merge up from master to get this, if possible

mogoodrich:
Got it, will do...

mogoodrich:
Unfortunately, changing the method signature from Double to Number leads to compile issues with the DrugOrderElement... :)  Since this is what you are working on, doesn't seem to make much sense for me to dive into this?

mseaton:
I'm not really working on that, and that should be trivial to fix, and actually that is probably what needed the fix when I initially threw it together.  The issue is that some DrugOrder properties are Integers (quantity, numRefills).  Whatever compilation errors you are seeing should be trivial to fix.  Probably just need "setInitialValue" to take in a Number rather than a Double, and to make sure when this widget is used it is cast to either Number or to Integer/Double appropriately.

mogoodrich:
I could dig into it tomorrow, but, no matter how trivial, it just seems silly for me to spend any time making changes to DrugOrderSubmissionElement (where the compile errors are) at this point since you are refactoring it?

mseaton:
I don't have changes to DrugOrderSubmissionElement - I'm creating a new tag.  I did identify the changes that I think should be needed to it (these are the only changes I have to DrugOrderSubmissionElement in my branch, which are all due to the changes I initially made to the NumberFieldWidget as described above):

1. Cast quantity to a Double:
`		orderTag.quantity = (Double) quantityWidget.getValue(session.getContext(), submission);
`
2. Cast duration to an Integer (can get rid of the null check)
`		orderTag.duration = (Integer) durationWidget.getValue(session.getContext(), submission);
`
3. Cast numRefills to an Integer (can get rid of the null check)
`		orderTag.numRefills = (Integer) numRefillsWidget.getValue(session.getContext(), submission);
`

mogoodrich:
After sleeping on this, I'm going to merge this in as-is... although the risk seems relatively low, the risk of refactoring and introducing method signature changes seems unnecessary just for the needs of this ticket (add a validation message)... definitely happy to consider refactoring as part of the overall Drug rework.

PR #210, opened 2020-10-29
HTML-758 - NumberFieldWidget should return Integer if floating point …

mseaton:
…is set to false


api/src/main/java/org/openmrs/module/htmlformentry/widget/NumberFieldWidget.java
```diff
@@ -207,24 +207,23 @@ private String getLocalizedErrorMessages() {
...
-	public Double getValue(FormEntryContext context, HttpServletRequest request) {
+	public Number getValue(FormEntryContext context, HttpServletRequest request) {
 		try {
-			Double d = (Double) HtmlFormEntryUtil.getParameterAsType(request, context.getFieldName(this), Double.class);
-			if (d != null && absoluteMinimum != null && d < absoluteMinimum) {
+			Number ret;
+			if (isFloatingPoint()) {
+				ret = (Double) HtmlFormEntryUtil.getParameterAsType(request, context.getFieldName(this), Double.class);
+			} else {
+				ret = (Integer) HtmlFormEntryUtil.getParameterAsType(request, context.getFieldName(this), Integer.class);
```
mogoodrich, line 216: Does this still properly render a "Not an Integer" error if the input is not a Integer?

mseaton: I don't think this has anything to do with that.  If you look at the generateHtml method, that seems to be all javascript driven on the number input on the client-side, and has nothing to do with pulling the value out of the request.

mogoodrich: It does, I specifically had to add this fix recently, see my description here:
https://issues.openmrs.org/browse/HTML-754

mseaton: @mogoodrich - thanks, fair enough.  Just pushed another commit with more error handling and some unit tests.

PR #211, opened 2020-10-29
Minor cleanup, including:

mseaton:
* log4j files should not be included in modules, as they overwrite the default openmrs appenders, this removes those in non-test projects
* CapturingPrintWriter is useful outside of a test context, so moving it
* A few private methods on DateWidget are useful to be able to use outside the class, make them public
* Remove unnecessary test dataset and just incorporate it into the main test dataset
* Add documentation for deploy web approaches in the pom
* Cleanup of incorrect javadoc
* Replace use of isVoided, which is deprecated and doesn't handle null, with BooleanUtils


mseaton:
I didn't create a ticket for this @mogoodrich - hopefully this is sufficiently non-controversial, especially considering this is part of the 4.0.0-SNAPSHOT release, which is a major version change.

PR #213, opened 2020-11-19
HTML-765 - Support initial values after they have been removed from t…

mseaton:
…he schema

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/SingleOptionWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/SingleOptionWidget.java
@@ -4,8 +4,14 @@ import javax.servlet.http.HttpServletRequest;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.openmrs.Concept;
+import org.openmrs.OpenmrsData;
+import org.openmrs.OpenmrsMetadata;
 import org.openmrs.OpenmrsObject;
 import org.openmrs.module.htmlformentry.FormEntryContext;
+import org.openmrs.module.htmlformentry.HtmlFormEntryUtil;
+import org.openmrs.module.htmlformentry.element.ValueStub;
+import org.openmrs.util.OpenmrsUtil;
 
 /**
  * This represents a single widget on a form which presents several coded options, of which only one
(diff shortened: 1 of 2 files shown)


mogoodrich (commented):
Generally looks good, but one comment:  should we also add a call to "addOptionIfMissing(this.initialValue)" if this.initialValue is not null to the setOptions method? (ie the use case where someone might set the initialValue before they set the options?)

mseaton:
It's a good question - I guess given how I implemented this in the end, that wouldn't hurt.  I considered this at the start and thought it might complicate things, and when I analyzed the code I found that in every instance of usage, options were fully set first and initialValue was set after, so I figured it would be ok.  If you'd like me to add that in I can.

mogoodrich:
Seems worth doing, just to be safe... but if there are cases I'm not thinking of where it might not do the right thing or actually hurt, let me know?

mseaton:
Thanks @mogoodrich  - done.

api/src/main/java/org/openmrs/module/htmlformentry/widget/SingleOptionWidget.java
```diff
@@ -36,6 +42,57 @@ public void setInitialValue(Object initialValue) {
...
+	}
+	
+	/**
+	 * The purpose of this method is to determine if the initial value configured on the widget is among
+	 * the options configured for the widget. If it is not among the options, then add it to the
+	 * available options, in order to ensure that existing forms that are opened and no longer contain
+	 * the same options that they were saved with are able to properly display and edit existing data
+	 * without loss.
+	 */
+	protected void addOptionIfMissing(Object initialValue) {
```
mks-d, line 56: I wonder if the original behaviour of `addOption(Option)` is still used as such in the end? From here on 'add option' will always be 'add option if missing' no?

Also I wonder why those `options` are not a set, based on an appropriate comparator, namely: `OpenmrsUtil.nullSafeEquals`.

mseaton: @mks-d that's a good point.  I've been going back and forth as to whether to act on this, but I'm going to leave it alone for now.

PR #214, opened 2020-11-23
HTML-755: Conversion from client to server timezone.

icrc-jfigueiredo:
[DRAFT] HTML-755 Convert from client timezone to server timezone before saving on DB.
This only works if the`<encounterDate>`  tag have ` showTime=true `

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
@@ -20,23 +20,7 @@ import java.lang.reflect.Method;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.GregorianCalendar;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.UUID;
+import java.util.*;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.BooleanUtils;
@@ -532,7 +516,7 @@ public class HtmlFormEntryUtil {
 	 * @param time the Date object that contains time information
 	 * @return a Date object with the combined date/time
 	 */
-	public static Date combineDateAndTime(Date date, Date time) {
+	public static Date combineDateAndTime(Date date, Date time, String timeZone) {
 		if (date == null)
 			return null;
 		Calendar cal = Calendar.getInstance();
@@ -545,6 +529,10 @@ public class HtmlFormEntryUtil {
 			cal.set(Calendar.MINUTE, temp.get(Calendar.MINUTE));
 			cal.set(Calendar.SECOND, temp.get(Calendar.SECOND));
 			cal.set(Calendar.MILLISECOND, temp.get(Calendar.MILLISECOND));
+			if (StringUtils.isNotEmpty(timeZone)) {
+				TimeZone tz = TimeZone.getTimeZone(timeZone);
+				cal.setTimeZone(tz);
+			}
 		}
 		return cal.getTime();
 	}
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
@@ -49,8 +49,15 @@ public class TimeWidget implements Widget {
 		if (context.getMode() == Mode.VIEW) {
 			String toPrint = "";
 			if (initialValue != null) {
+				StringBuilder sb = new StringBuilder();
 				toPrint = timeFormat().format(initialValue);
-				return WidgetFactory.displayValue(toPrint);
+				sb.append(WidgetFactory.displayValue(toPrint));
+				//Format the datetime with server timezone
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
```diff
@@ -49,8 +49,14 @@ public String generateHtml(FormEntryContext context) {
...
 			String toPrint = "";
 			if (initialValue != null) {
+				StringBuilder sb = new StringBuilder();
 				toPrint = timeFormat().format(initialValue);
-				return WidgetFactory.displayValue(toPrint);
+				//Get server timezone
+				SimpleDateFormat dateFormatYouWant = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+				String newDate = dateFormatYouWant.format(initialValue);
+				String timezone = newDate.substring(newDate.length()-5);
+				sb.append("<span class=\"value\" tz=\"").append(timezone).append("\">").append(toPrint).append("</span>");
```
mks-d, line 58: This leaves a possibility for the frontend to convert back to the client timezone in VIEW mode, correct?

Btw you can simply do:
```java
String timezone = new SimpleDateFormat("Z").format(initialValue);
```

api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
```diff
@@ -124,6 +130,9 @@ public String generateHtml(FormEntryContext context) {
 					}
 					sb.append("</select>");
 				}
+				sb.append("<input type=\"hidden\" class=\"hfe-timeZone\" name=\"").append(context.getFieldName(this))
+						.append("timeZone").append("\">");
+				sb.append("</input>");
```
mks-d, line 135: And this creates a possibility for the frontend to submit the timezone in ENTER and EDIT mode, correct?

api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
```diff
@@ -134,6 +143,18 @@ public String generateHtml(FormEntryContext context) {
...
+	public String getTimeZone(FormEntryContext context, HttpServletRequest request) {
+		try {
+			String tz = (String) HtmlFormEntryUtil.getParameterAsType(request, context.getFieldName(this) + "timeZone",
+					String.class);
+			return tz;
+		}
+		catch (Exception ex) {
+			throw new IllegalArgumentException("Illegal value");
+		}
+	}
```
mks-d, line 156: Change this to slightly simplified and renamed:
```java
/**
 * @return The timezone string info that was submitted as part of the time submission. 
 */
public String getSubmittedTimeZone(FormEntryContext context, HttpServletRequest request) {
    return (String) HtmlFormEntryUtil.getParameterAsType(request, context.getFieldName(this) + "timeZone", String.class);
}
```

api/src/test/java/org/openmrs/module/htmlformentry/widget/TimeWidgetTest.java
```diff
@@ -60,7 +61,7 @@ protected void testTimeFormat(TimeWidget widget, String expectedValue) {
 		when(formEntryContext.getMode()).thenReturn(FormEntryContext.Mode.VIEW);
 		String html = widget.generateHtml(formEntryContext);
 		String expected = "<span class=\"value\">" + expectedValue + "</span>";
-		assertEquals(expected, html);
+		assertTrue(StringUtils.containsIgnoreCase(html, expected));
```
mks-d, line 64: I don't think you need to ignore case anymore.

mks-d: We could do this:
```java
protected void testTimeFormat(TimeWidget widget, String expectedValue) {
    FormEntryContext formEntryContext = mock(FormEntryContext.class);
    when(formEntryContext.getMode()).thenReturn(FormEntryContext.Mode.VIEW);
    String html = widget.generateHtml(formEntryContext);
    String timezone = new SimpleDateFormat("Z").format(new Date());
    String expected = "<span class=\"value\" tz=\"" + timezone + "\">" + expectedValue + "</span>";
    assertEquals(expected, html);
}
```

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -532,7 +516,7 @@ public static HtmlFormSchema getHtmlFormSchema(HtmlForm htmlForm, Mode mode) {
 	 * @param time the Date object that contains time information
 	 * @return a Date object with the combined date/time
 	 */
-	public static Date combineDateAndTime(Date date, Date time) {
+	public static Date combineDateAndTime(Date date, Date time, String timeZone) {
```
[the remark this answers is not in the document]
icrc-jfigueiredo: Every place on this project that we call that function i changed the parameters to send a timezone (or null if they dont have one), we should not have a problem once they accept this changes, correct?

api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
icrc-jfigueiredo: Yes, correct.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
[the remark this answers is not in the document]
mks-d: Other modules may rely on this API. HFE UI for instance, but even any other module that depends on HFE for whatever reason.
As a general rule, it is good practise to maintain existing APIs until they **really** need to be deprecated.

mks-d:
@icrc-jfigueiredo you can also just force push here if there are too many changes to reconcile.
... or open a new PR, up to you.

PR #216, opened 2020-12-01
HTML-761: Obs tag with answerClasses to filter out retired concepts.

sherrif10:
https://issues.openmrs.org/browse/HTML-761
cc @mks-d  @ibacher


api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -698,7 +698,8 @@ private void prepareWidgets(FormEntryContext context, Map<String, String> parame
 							if (cc == null) {
 								throw new RuntimeException("Cannot find concept class " + className);
 							}
-							conceptAnswers.addAll(Context.getConceptService().getConceptsByClass(cc));
+							conceptAnswers.addAll((Collection<? extends Concept>) Context.getConceptService()
+							        .getConceptsByClass(cc).stream().filter(it -> !it.getRetired()));
```
ibacher, line 702:
```suggestion
							conceptAnswers.addAll(Context.getConceptService()
							        .getConceptsByClass(cc).stream().filter(it -> !it.getRetired())
							        .collect(Collectors.toList()));
```

sherrif10: Resolved thanks

mogoodrich (commented):
LGTM... but wondering if @mseaton  has any thoughts on this as he recently did a (somewhat) related change to HFE.

sherrif10:
Thanks @mogoodrich

mseaton (approved):
The PR that I'm working on for HTML-765 _should_ ensure that previously saved values will still appear in edit mode, so I don't think this will adversely affect that use case, assuming I get that in.

I'm not 100% sure I agree this fix is needed, but I don't see any harm in it I guess.  One could argue that if a Concept is explicitly added as a valid Answer to a Concept, then it should be displayed, and if you don't want it to be listed any more, just remove it from the allowed answers to a given Question.  @mks-d - looks like you created this ticket, so I'll defer to you on whether you've given that particular thought.

sherrif10:
Thanks @mseaton for clarifying, So the idea was that if any concept is retired , it shouldnot reflect again. cc @mks-d

mogoodrich:
@mseaton  I think this only applies to the "answerClasses" attribute, which displays all concepts that depend to a class or set of classes...

sherrif10:
@mogoodrich  Yeah sure basically coded concepts with answerClasses Attribute, so the idea is that Obs tag should not return retired coded concept

sherrif10:
Hello @mks-d  Any more suggestion on this work being done here thanks

sherrif10:
Shall we have a merge into master because i think its ready cc @mks-d

sherrif10 (commented):
LGTM  this can be merged cc @mksd @dkayiwa  thnks

mks-d (changes requested):
@mseaton about your comment [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/216#pullrequestreview-543344284), as @mogoodrich said the scope here is only about the `answerClasses` attribute.

LGTM except for the **.gitignore** file that should be untouched.

api-2.3/.gitignore
```diff
@@ -0,0 +1 @@
+/target/
```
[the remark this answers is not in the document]
sherrif10: ohh sure let me fix it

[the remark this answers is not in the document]
sherrif10: Resolved thanks

api-2.3/.gitignore
```diff
@@ -0,0 +1 @@
+
```
mks-d, line 1: Still there.

mks-d (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -698,7 +699,8 @@ private void prepareWidgets(FormEntryContext context, Map<String, String> parame
 							if (cc == null) {
 								throw new RuntimeException("Cannot find concept class " + className);
 							}
-							conceptAnswers.addAll(Context.getConceptService().getConceptsByClass(cc));
+							conceptAnswers.addAll(Context.getConceptService().getConceptsByClass(cc).stream()
+							        .filter(it -> !it.getRetired()).collect(Collectors.toList()));
```
mks-d, line 703: A last detail, in `it -> !it.getRetired()` just change the variable name from `it` to `c` ("c" for "concept"). More readable.

mks-d (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
sherrif10: Thanks @mks-d  have responded to the comment  above

mks-d (approved):

PR #217, opened 2020-12-03
HTML-765 - Followup fix to EncounterDetailSubmissionElement and Singl…

mseaton:
…eOptionWidget to ensure the right set of options are configured given initial and default values for encounter location and provider widgets.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
@@ -198,62 +198,41 @@ public class EncounterDetailSubmissionElement implements HtmlGeneratorElement, F
 						}
 					}
 				}
-				
-				// Otherwise, use default options appropriate to the underlying OpenMRS version
+				// Otherwise, use default options
 				else {
-					if (openmrsVersionDoesNotSupportProviders()) {
-						// limit to users with the default OpenMRS PROVIDER role,
-						String defaultRole = RoleConstants.PROVIDER;
-						Role role = Context.getUserService().getRole(defaultRole);
-						if (role != null) {
-							users = Context.getService(HtmlFormEntryService.class).getUsersAsPersonStubs(role.getRole());
-						}
-						// If this role isn't used, default to all Users
-						if (users.isEmpty()) {
-							users = Context.getService(HtmlFormEntryService.class).getUsersAsPersonStubs(null);
-						}
-					} else {
-						// in OpenMRS 1.9+, get all suitable providers
-						users = getAllProvidersThatArePersonsAsPersonStubs();
-					}
+					users = getAllProvidersThatArePersonsAsPersonStubs();
 				}
 				
 				for (PersonStub personStub : users) {
-					
-					Option option = new Option(StringEscapeUtils.escapeHtml(personStub.toString()),
-					        personStub.getId().toString(), false);
+					String optionLabel = StringEscapeUtils.escapeHtml(personStub.toString());
+					Option option = new Option(optionLabel, personStub.getId().toString(), false);
 					providerUsers.add(option);
 				}
 				providerOptions.addAll(providerUsers);
-				
 			}
 			
-			// Set default values as appropriate
-			Person defaultProvider = null;
-			Option defProviderOption;
-			if (context.getExistingEncounter() != null) {
-				defaultProvider = EncounterCompatibility.getProvider(context.getExistingEncounter());
-				// this is done to avoid default provider being added twice due to that it can be added from the
-				// users = getAllProvidersThatArePersonsAsPersonStubs(); section with selected="false", therefore this can't be caught when
-				// searching whether the options list contains the 'defaultProvider'
-				boolean defaultOptionPresent = false;
-				if (defaultProvider != null) {
-					for (Option option : providerOptions) {
-						if (option.getValue().equals(defaultProvider.getId().toString())) {
-							defaultOptionPresent = true;
-							providerOptions.remove(option);
-							break;
-						}
-					}
-				}
-				if (defaultOptionPresent) {
-					defProviderOption = new Option(
-					        StringEscapeUtils.escapeHtml(defaultProvider.getPersonName().getFullName()),
-					        defaultProvider.getId().toString(), true);
-					providerOptions.add(defProviderOption);
-				}
-				
+			Collections.sort(providerOptions, new OptionComparator());
+			
+			if (("autocomplete").equals(parameters.get("type"))) {
+				providerWidget.addOption(new Option());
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -198,62 +198,41 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
 						}
 					}
 				}
-				
-				// Otherwise, use default options appropriate to the underlying OpenMRS version
+				// Otherwise, use default options
 				else {
-					if (openmrsVersionDoesNotSupportProviders()) {
```
mseaton, line 204: We don't support these old versions any more, so I just removed these switch blocks while I was in here to make things cleaner

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -198,62 +198,41 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
...
-						users = getAllProvidersThatArePersonsAsPersonStubs();
-					}
+					users = getAllProvidersThatArePersonsAsPersonStubs();
 				}
 				
 				for (PersonStub personStub : users) {
-					
-					Option option = new Option(StringEscapeUtils.escapeHtml(personStub.toString()),
-					        personStub.getId().toString(), false);
+					String optionLabel = StringEscapeUtils.escapeHtml(personStub.toString());
```
mseaton, line 207: Non functional, just moving things to fit on lines better

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -198,62 +198,41 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
...
 				}
 				providerOptions.addAll(providerUsers);
-				
 			}
 			
-			// Set default values as appropriate
-			Person defaultProvider = null;
-			Option defProviderOption;
-			if (context.getExistingEncounter() != null) {
-				defaultProvider = EncounterCompatibility.getProvider(context.getExistingEncounter());
```
mseaton, line 235: All of the code that deals with setting default values I have moved to the end of the section after all of the options are configured on the widget.

All of the removed code below here I refactored.  Whereas previously there was a lot of adding and removing options to avoid duplicates, now the addOption method + setInitialValue handle this and all of this is unnecessary.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -263,44 +242,26 @@ public EncounterDetailSubmissionElement(FormEntryContext context, Map<String, Ob
...
-				providerWidget
-				        .addOption(new Option(Context.getMessageSourceService().getMessage("htmlformentry.chooseAProvider"),
-				                "", !initialValueIsSet)); // if no initial or default value
-				
-				if (!providerOptions.isEmpty()) {
-					for (Option option : providerOptions) {
-						providerWidget.addOption(option);
-					}
-					
+			// Set the selected provider option appropriately
```
mseaton, line 260: Above, all of the options are added with selected = false.  This simply runs through them all and sets selected based on the initial value that was set.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -422,30 +383,10 @@ else if (parameters.get("order") != null) {
 				locations = removeLocationsNotEqualToOrDescendentOf(locations, ((Visit) context.getVisit()).getLocation());
 			}
 			
-			// Set default values
```
mseaton, line 425: Like the provider section, determining and setting initial/default values is moved later, after the options have been added to the widget

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -422,30 +383,10 @@ else if (parameters.get("order") != null) {
...
-				if (!locations.contains(defaultLocation)) {
-					locations.add(defaultLocation);
-				}
-			}
-			
 			// now create the actual location options
 			for (Location location : locations) {
 				String label = HtmlFormEntryUtil.format(location);
-				Option option = new Option(label, location.getId().toString(), location.equals(defaultLocation));
+				Option option = new Option(label, location.getId().toString(), false);
```
mseaton, line 389: We don't have this information here, so we add the options with selected = false.  After everything is configured in a later section we iterate over the options and set selected appropriately

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -461,12 +402,45 @@ else if (parameters.get("order") != null) {
...
 				if (!locationOptions.isEmpty()) {
-					for (Option option : locationOptions)
+					for (Option option : locationOptions) {
 						locationWidget.addOption(option);
+					}
+				}
+			}
+			
+			// Configure the initial value of the widget with any location value from an existing encounter
+			Location initialValue = null;
```
mseaton, line 415: We distinguish an "initialValue" from a "defaultValue".  An initialValue is one we want to configure and we want it to add an option if it is missing.  A defaultValue is one that we want to configure as the default selected value if the value is one of the existing options.  This is why these are now kept distinct.

api/src/main/java/org/openmrs/module/htmlformentry/widget/SingleOptionWidget.java
```diff
@@ -56,17 +56,7 @@ public void setInitialValue(Object initialValue) {
...
-				for (Option option : options) {
-					if (OpenmrsUtil.nullSafeEquals(option.getValue(), initialValueOption.getValue())) {
-						optionFound = true;
-					}
-				}
-			}
-			if (!optionFound) {
-				addOption(initialValueOption);
-			}
+			addOption(initialValueOption);
```
mseaton, line 59: This is in line with the comment that @mks-d made in the earlier PR / commit for this ticket.  All "addOption" should really be "addOptionIfMissing", since we never want to end up with duplicate options in the list.  So this logic is moved into the addOption method

mogoodrich (approved):
LGTM, thanks @mseaton !  I'm going to go ahead and merge to fix the build...

PR #219, opened 2020-12-15
HTML-755: <encounterDate showTime="true"/> to support time zones.

icrc-jfigueiredo:
#### Ticket
https://issues.openmrs.org/browse/HTML-755

#### What I have done
Converts from client timezone to server timezone.
Also systematically converts form server timezone to UTC when sending the dates to the client.


api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -24,33 +24,16 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.time.ZonedDateTime;
+import java.util.*;
```
mks-d, line 36: Avoid wildcard imports.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -96,6 +77,28 @@ private MetadataMappingResolver getMetadataMappingResolver() {
 		return Context.getRegisteredComponent("metadataMappingResolver", MetadataMappingResolver.class);
 	}
 	
+	private void RegisterDateAndTime(FormEntryContext context, DateWidget widget, Map<String, Object> parameters) {
```
mks-d, line 80: The naming is most likely wrong (and also method names are always `camelCase`'d in Java).

Is this taking an empty `DateWidget` instance to initialize it from the context and the params?

If yes change the signature to
```java
private void initializeDateWidget(FormEntryContext context, Map<String, Object> parameters, DateWidget widget) {
...
```
With inputs as first args (context and params) and output as last arg (the date widget).

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -604,6 +574,13 @@ private Boolean isLocationEqualToOrDescendentOf(Location location, Location test
 		}
 	}
 	
+	public void generateHtmlDateWidget(FormEntryContext context, DateWidget widget) {
+		id = id != null ? id : "encounterDate";
+		context.registerPropertyAccessorInfo(id + ".value", context.getFieldNameIfRegistered(widget),
+		    "dateFieldGetterFunction", null, "dateSetterFunction");
+		context.registerPropertyAccessorInfo(id + ".error", context.getFieldNameIfRegistered(widget), null, null, null);
+	}
```
mks-d, line 582: What's this method doing exactly? Does it need to be public?

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -715,6 +687,14 @@ private String getSetterFunction(Widget widget) {
 		return null;
 	}
 	
+	public void validateWidget(FormEntryContext context, DateWidget widget, HttpServletRequest submission) throws Exception {
+		Date date = (Date) widget.getValue(context, submission);
+		if (date == null)
+			throw new Exception("htmlformentry.error.required");
+		if (OpenmrsUtil.compare((Date) date, new Date()) > 0)
+			throw new Exception("htmlformentry.error.cannotBeInFuture");
```
mks-d, line 695: Always embrace `if`s content:
```java
if (date == null) {
  ...
}
```

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,252 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.TimeZone;
+
+public class ZonedDateTimeWidget extends DateWidget implements Widget {
+	
+	public static final String DEFAULT_TIME_FORMAT = "HH:mm";
+	
+	protected Date initialValue;
```
mks-d, line 19: This member is already defined [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/219/files#diff-9a69cc35cad580005d556de11ba79dffaa7873f7b3efaf957a094fb38b61355bR21) in `DateWidget`.

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,252 @@
...
+	
+	protected boolean hidden;
```
mks-d, line 21: Same remark.

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,252 @@
...
+						sb.append("<option value=\"" + i + "\"");
+						if (valAsCal != null) {
+							if (valAsCal.get(Calendar.SECOND) == i)
+								sb.append(" selected=\"true\"");
+						}
+						sb.append(">" + label + "</option>");
+					}
+					sb.append("</select>");
+				}
+				sb.append("<input type=\"hidden\" class=\"hfe-timeZone\" name=\"").append(context.getFieldName(this))
```
mks-d, line 157: A detail but let's systematically write time zone (in code) as `timezone`, all lower caps.
So this would become `class=\"hfe-timezone\"`.

api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java
```diff
@@ -18,18 +18,18 @@
  */
 public class DateWidget implements Widget {
 	
-	private Date initialValue;
+	protected Date initialValue;
```
mks-d, line 21: You shouldn't need to make all those guys `protected` since you have public getters and setters at the level of `DateWidget`.

mks-d (changes requested):

api/src/main/resources/messages.properties
```diff
@@ -127,7 +127,7 @@ htmlformentry.warning.invalidAnswerConcept              = The answer concept spe
 htmlformentry.chooseAnEncounterRole=Choose an Encounter Role
 htmlformentry.unknownProviderName=Unknown
 htmlformentry.providerPlaceHolder=Type identifier or name
-
+htmlformentry.error.handleTimezones=To use timezones the server need to have the variable handleTimezones as true.
```
mks-d, line 130: This isn't used anywhere is it?

omod/src/main/resources/config.xml
```diff
@@ -98,7 +98,12 @@
         <description>When set to true and an encounter is being entered in the context of a visit, the encounter location will be restricted to the current visit location (and it's descendents)</description>
     </globalProperty>
 	<!-- /Required Global Properties -->
-
+	<globalProperty>
+		<property>htmlformentry.handleTimezones</property>
+		<defaultValue>false</defaultValue>
+		<description>Global variable to allow the sistem to user timezones, to use it, it should be true, but for retrocompatibility it´s initialized as false, besides that var, to use timezones the "encounterdate" tag should have "showTime=true".
+		</description>
+	</globalProperty>
```
mks-d, line 106: And this isn't used either?

mks-d (commented):
I'll make a PR that starts reorganising the code in a way that will make it easier to make `ZonedDateTimeWidget.generateHtml(FormEntryContext)` closer to being build from reused pieces of both `DateWidget` and `TimeWidget`.

For now my main issue is that the VIEW mode turns the datetime into UCT while the EDIT mode doesn't.
That surely isn't ok, what was the thinking there?

api/src/main/resources/messages.properties
icrc-jfigueiredo: > I'll make a PR that starts reorganising the code in a way that will make it easier to make `ZonedDateTimeWidget.generateHtml(FormEntryContext)` closer to being build from reused pieces of both `DateWidget` and `TimeWidget`.
> 
> For now my main issue is that the VIEW mode turns the datetime into UCT while the EDIT mode doesn't.
> That surely isn't ok, what was the thinking there?

The Enter/Edit mode get their dates from [HFE-UI](https://github.com/icrc-jfigueiredo/openmrs-module-htmlformentryui/commit/e6f08405d341c057f976546e91728be6c453195a#diff-9920bfbf71a8b9f88051301b5ce589794fa2de063329605ce35381698ee79285) the change to UTC is done there.

omod/src/main/resources/config.xml
icrc-jfigueiredo: > And this isn't used either?

This is used to set the initial value of the GP to false. In this version i am already using the GP, here and here.
For testing that GP I am setting it to true [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/219/commits/fa994910c7793ff4562a80bae8574924fc3bbdf6#diff-4fd897c96b3194ac36e52c2765e808d6be3202d2adb2a18780c41fec03b635e0R727) and checking it's value [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/219/commits/fa994910c7793ff4562a80bae8574924fc3bbdf6#diff-4fd897c96b3194ac36e52c2765e808d6be3202d2adb2a18780c41fec03b635e0R738).

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,104 @@
...
+		dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+		return dateFormat;
+	}
+	
+	public String generateHtml(FormEntryContext context) {
+		
+		if (context.getMode() == FormEntryContext.Mode.VIEW) {
+			if (initialValue != null) {
+				return new StringBuilder().append("<span id=\"dateTimeWithTimezone\" class=\"value\">")
+				        .append(getUTCDateFormat().format(initialValue)).append("</span>").toString();
```
mks-d, line 43: @icrc-jfigueiredo this is problematic, HTML ids should be unique, why couldn't you just do
```java
return WidgetFactory.displayValue(toPrint);
```
as done [here](https://github.com/openmrs/openmrs-module-htmlformentry/blob/a2d606bfb463b70671f4d5c2fadfcf9b77904c85/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java#L51-L54) or [here](https://github.com/openmrs/openmrs-module-htmlformentry/blob/a2d606bfb463b70671f4d5c2fadfcf9b77904c85/api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java#L80-L83)?

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -715,24 +696,50 @@ private String getSetterFunction(Widget widget) {
...
+	// ------------------FOR TESTING TZ-----------------------//
 	/**
 	 * @see FormSubmissionControllerAction#validateSubmission(FormEntryContext, HttpServletRequest)
 	 */
 	@Override
 	public Collection<FormSubmissionError> validateSubmission(FormEntryContext context, HttpServletRequest submission) {
 		List<FormSubmissionError> ret = new ArrayList<FormSubmissionError>();
+		// ------------------FOR TESTING TZ-----------------------//
+		//setGlobalProperty(HtmlFormEntryConstants.GP_HANDLE_TIMEZONES, "true");
+		// ------------------FOR TESTING TZ-----------------------//
```
mks-d, line 728: @icrc-jfigueiredo let's go through a cleanup of `EncounterDetailSubmissionElement` now, and in particular that removes the "code for testing" that will not stay.

mks-d (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
icrc-jfigueiredo: In this case we use HTML because we want to have a specific ID, so i can see if that ID is present where we execute the client JS, and if that ID is present we convert that date from UTC-> Client TZ.
So we use that html to generate an ID and check if we need to convert that timezone.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -730,7 +732,7 @@ protected void validateDateWidget(FormEntryContext context, DateWidget dateWidge
 				String serverTimezone = TimeZone.getDefault().getID();
 				boolean handleTimezones = Boolean.parseBoolean(
 				    Context.getAdministrationService().getGlobalProperty(HtmlFormEntryConstants.GP_HANDLE_TIMEZONES));
-				if (!serverTimezone.equals(clientTimezone) && !handleTimezones) {
+				if (!serverTimezone.equals(clientTimezone) && !handleTimezones && (clientTimezone != null)) {
```
mks-d, line 735: @icrc-jfigueiredo my thinking here is that `serverTimezone` will never be blank because `TimeZone.getDefault().getID()` should always give _something_.
So if `clientTimezone` is `null`, then this condition is enough: `!serverTimezone.equals(clientTimezone)`.

Am I missing something?

icrc-jfigueiredo: You are correct, I was testing some scenarios.

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,103 @@
...
+		dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+		return dateFormat;
+	}
+	
+	public String generateHtml(FormEntryContext context) {
+		
+		if (context.getMode() == FormEntryContext.Mode.VIEW) {
+			if (initialValue != null) {
+				String toPrint = getUTCDateFormat().format(initialValue);
+				return WidgetFactory.displayValue(toPrint, "dateTimeWithTimezone");
```
mks-d, line 43: @icrc-jfigueiredo so this means that the client side code is responsible for turning this RFC 3339 date into what the user is supposed to see (which is indeed one of the the new conventions about time zones and OpenMRS). I guess that happens in HFE UI right?

icrc-jfigueiredo: Correct, https://github.com/openmrs/openmrs-module-htmlformentryui/compare/master...icrc-jfigueiredo:HTML-755#diff-7543ad74c606f05f56deebe2a2c1eb2ba0cb06eb6de7400957dfe8824939eceeR3

mks-d: Let's rename this class `"date-with-timezone"`, ref. https://csswizardry.com/2010/12/css-camel-case-seriously-sucks/

mks-d (changes requested):

api-tests/src/test/java/org/openmrs/module/htmlformentry/RegressionTest.java
```diff
@@ -2514,6 +2515,7 @@ public void setupRequest(MockHttpServletRequest request, Map<String, String> wid
 				request.addParameter("w1hours", String.valueOf(cal.get(Calendar.HOUR_OF_DAY)));
 				request.addParameter("w1minutes", String.valueOf(cal.get(Calendar.MINUTE)));
 				request.addParameter("w1seconds", String.valueOf(cal.get(Calendar.SECOND)));
+				request.setParameter("w1timezone", TimeZone.getDefault().getID()); // making sure that client tz = server tz
```
mks-d, line 2518: @mogoodrich actually this is where we see some non-backward compatible stuff. The new behaviour of `<encounterDate showTime="true"/>` implies that a `timezone` parameter must be submitted or an an exception is thrown (exception that didn't exist before).

mks-d (commented):
@mogoodrich I have done a walk through set of comments for you...

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -120,6 +125,20 @@
...
+	public static String toRFC3339(Date date) {
+		return ISODateTimeFormat.dateTime().print(new DateTime(date.getTime(), UTC));
+	}
+	
+	/**
+	 * @return The GregorianCalendar set in UTC for the date.
+	 */
+	public static Calendar toUTCCalendar(Date date) {
+		return new DateTime(date.getTime(), UTC).toGregorianCalendar();
+	}
```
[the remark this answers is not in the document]
dkayiwa: How about putting them in openmrs core and we back port? Then a module like this could have a comment with instructions to remove this duplicated code once we upgrade to the openmrs core version with these new methods.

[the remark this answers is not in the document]
mks-d: @dkayiwa that's 🎼 to my ears. Makes sense.
I'll need to do a little roundtable with people so that we agree on which methods that should be those Core time zone utils, but otherwise that's the idea.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryConstants.java
```diff
@@ -36,9 +36,12 @@
 	
 	public static final String GP_RESTRICT_ENCOUNTER_LOCATION_TO_CURRENT_VISIT_LOCATION = "htmlformentry.restrictEncounterLocationToCurrentVisitLocation";
 	
+	public final static String GP_HANDLE_TIMEZONES = "htmlformentry.handleTimezones";
```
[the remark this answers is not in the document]
mogoodrich: Okay... I will only review the encounter date stuff then, I may have further comments on follow-on commits?

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -120,6 +125,20 @@
 	
 	public static Log log = LogFactory.getLog(HtmlFormEntryUtil.class);
 	
+	/**
+	 * @return The date formated as RFC 3339.
+	 */
+	public static String toRFC3339(Date date) {
+		return ISODateTimeFormat.dateTime().print(new DateTime(date.getTime(), UTC));
```
mogoodrich, line 132: Just to confirm, this is the same format that our REST endpoints use?

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -715,24 +705,36 @@ private String getSetterFunction(Widget widget) {
...
-					Date time = (Date) timeWidget.getValue(context, submission);
-					date = HtmlFormEntryUtil.combineDateAndTime(date, time);
+				validateDateWidget(context, dateWidget, submission);
+			} else if (zonedDateTimeWidget != null) {
+				String clientTimezone = zonedDateTimeWidget.getSubmittedTimezone(context, submission);
+				String serverTimezone = TimeZone.getDefault().getID();
+				boolean handleTimezones = Boolean.parseBoolean(
+				    Context.getAdministrationService().getGlobalProperty(HtmlFormEntryConstants.GP_HANDLE_TIMEZONES));
+				if (!serverTimezone.equals(clientTimezone) && !handleTimezones) {
+					throw new IllegalStateException("htmlformentry.error.handleTimezones");
```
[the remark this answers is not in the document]
mogoodrich: Makes sense to me!

omod/src/main/resources/config.xml
```diff
@@ -98,7 +98,12 @@
         <description>When set to true and an encounter is being entered in the context of a visit, the encounter location will be restricted to the current visit location (and it's descendents)</description>
     </globalProperty>
 	<!-- /Required Global Properties -->
-
+	<globalProperty>
+		<property>htmlformentry.handleTimezones</property>
+		<defaultValue>true</defaultValue>
+		<description>When set to true dates are sent from the server to the client as UTC dates and parsed from the client to the server as holding the client timezone information.
+		</description>
```
mogoodrich, line 105: This isn't necessarily true, correct?  If you have got an encounterDate with showTime='false' it will still revert to the old behavior even if this GP is set to true?  (

mogoodrich (approved):
Had a few comments, but generally looks good to me.

I'm generally happy that this doesn't seem to to touch *too* much of the code yet, though that may change as you apply it to other widgets.

I'll be happy to run this through our smoke tests and do some manual testing once this goes though, thanks!

omod/src/main/resources/config.xml
icrc-jfigueiredo: Correct, for now, to be able to use timezones we need to have showTime='true' AND the GP true or the serer timezone be the same as the client timezone.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
mks-d: @mogoodrich yes.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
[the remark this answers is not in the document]
mks-d: @dkayiwa @mseaton should we try to match what Core will provide?
Like having such package: `org.openmrs.util.TimeZoneUtil`?

omod/src/main/resources/config.xml
mks-d: @mogoodrich, what about:

>When set to true dates are sent from the server to the client as UTC dates, and time-sensitive widgets send dates from the client to the server as holding the client time zone information.

?

`<encounterDate showTime="false"/>` would be considered as a non time-sensitive widget (whereas `<encounterDate showTime="true"/>` is.)

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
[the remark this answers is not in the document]
dkayiwa: Looks like that has been the practice for things we intend to eventually move into core. That way, module developers do not have to change the package name.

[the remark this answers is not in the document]
mks-d: Ok. @dkayiwa any existing util class that we should add those methods to in your opinion? Rather than adding yet another util class.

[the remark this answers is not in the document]
dkayiwa: Though the `OpenmrsUtil` class already has some date utilities, i find it bloated with too many responsibilities. I therefore find it cleaner to use a new utility class.

omod/src/main/resources/config.xml
mogoodrich: I *think* I get it... :)

api/src/main/java/org/openmrs/module/htmlformentry/widget/ZonedDateTimeWidget.java
```diff
@@ -0,0 +1,105 @@
...
+		DateFormat dateFormat = new SimpleDateFormat("dd/MM/yyyy");
+		dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+		return dateFormat;
+	}
+	
+	public String generateHtml(FormEntryContext context) {
+		
+		if (context.getMode() == FormEntryContext.Mode.VIEW) {
+			if (initialValue != null) {
+				return WidgetFactory.displayValue(toRFC3339(initialValue), "rfc3339-date");
```
mks-d, line 40: @mogoodrich just bringing your attention to this as well. In VIEW mode dates as rendered as RFC 3339 (as agreed) and we have tagged them with the CSS class `.rfc3339-date` so that there's an easy marker for any client side code to localize them.

mogoodrich: So you are kind of relying on there being some client-side code to render it in a human-readable format?  (that's fine with me, just confirming)

mks-d: I will go as far as saying that it's (one of) our new conventions: https://wiki.openmrs.org/display/docs/Time+Zones+Conventions#TimeZonesConventions-ClientDisplay:InLocalTimeZone :)

mogoodrich: The convention just seems sort of weird in this old model (where the server is rendering what was originally assumed to be simple display text, not actual objects)... but, again, I don't see a way around it, so I'm okay with it.

mks-d: Sure, you mean old school HFE predating UIFR?

As @bmamlin said ([here](https://talk.openmrs.org/t/hfe-better-handling-of-timezones/31659/26?u=mksd)) the convention is a guide to the best practise, not a mandatory rule:

>The goal of conventions is not to make unbreakable rules; rather, to define & promote best (preferred) practice and make it easier to identify when we are breaking our conventions (so we have to justify breaking them).

omod/src/main/resources/config.xml
```diff
@@ -99,7 +99,7 @@
     </globalProperty>
 	<!-- /Required Global Properties -->
 	<globalProperty>
-		<property>htmlformentry.handleTimezones</property>
+		<property>uiframework.timezone.conversions</property>
```
mks-d, line 102: Same here.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryConstants.java
```diff
@@ -36,7 +36,7 @@
 	
 	public static final String GP_RESTRICT_ENCOUNTER_LOCATION_TO_CURRENT_VISIT_LOCATION = "htmlformentry.restrictEncounterLocationToCurrentVisitLocation";
 	
-	public final static String GP_HANDLE_TIMEZONES = "htmlformentry.handleTimezones";
+	public final static String GP_TIMEZONE_CONVERSIONS = "uiframework.timezone.conversions";
```
mks-d, line 39: This will have to be
```java
"htmlformentry.timezone.conversions"
```
HFE can be used outside of UIFR and cannot refer to (or take care of creating) a UIFR-specific GP.

However for UIFR-dependent modules, only the UIFR GP will be checked. So in practise `"uiframework.timezone.conversions"` will superseed `"htmlformentry.timezone.conversions"`.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -734,7 +734,7 @@ protected void validateDateWidget(FormEntryContext context, DateWidget dateWidge
 				
 				String serverTimezone = TimeZone.getDefault().getID();
 				boolean handleTimezones = Boolean.parseBoolean(
```
mks-d, line 736: Maybe `handleTimezones` → `convertTimezones`?

mks-d (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
icrc-jfigueiredo: Sure.

icrc-jfigueiredo: But in that case, maybe we should also change [this ](https://github.com/openmrs/openmrs-module-uiframework/compare/master...icrc-jfigueiredo:HTML-755-2#diff-6e177127ee5b17a0d0a088acefa4c4d980d03a506959223081b579f153e8382fR629)on UiFramework UiUtils.

mks-d (approved):

mogoodrich:
I know I approved this a few months ago, I see @mks-d  just approved it... any changes I should re-review?

icrc-jfigueiredo:
Hi @mogoodrich the last changes we made, were mostly about changing names. But please feel free to review it. Thank you

icrc-jfigueiredo:
Hi @mogoodrich , what do you think about removing [this ](https://github.com/icrc-jfigueiredo/openmrs-module-htmlformentry/blob/HTML-755/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java#L164) condition:
`if (h == 0 && m == 0 && s == 0) {
				return null;
			}` , this is in TimeWidget, with this condition it is not possible to save a form at midnight (00:00:00).

mogoodrich:
> Hi @mogoodrich , what do you think about removing [this ](https://github.com/icrc-jfigueiredo/openmrs-module-htmlformentry/blob/HTML-755/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java#L164) condition:
> `if (h == 0 && m == 0 && s == 0) { return null; }` , this is in TimeWidget, with this condition it is not possible to save a form at midnight (00:00:00).
> CC:@mks-d

Seem fine/correct to me!

api/src/main/resources/messages.properties
```diff
@@ -153,19 +153,21 @@ htmlformentry.error.notGreaterThan                      = Cannot be greater than
 htmlformentry.error.mustBeAtLeast                       = Must be at least
 htmlformentry.warning.invalidAnswerConcept              = The answer concept specified is not a valid concept for this question concept
 
-htmlformentry.chooseAnEncounterRole=Choose an Encounter Role
-htmlformentry.unknownProviderName=Unknown
-htmlformentry.providerPlaceHolder=Type identifier or name
+htmlformentry.chooseAnEncounterRole                     = Choose an Encounter Role
```
dkayiwa, line 156: Did you add these spaces by accident?

api/src/main/resources/messages.properties
```diff
@@ -153,19 +153,21 @@ htmlformentry.error.notGreaterThan                      = Cannot be greater than
...
-htmlformentry.conditionui.additionalDetail.label=Additional Detail
+htmlformentry.error.handleTimezones                     = The attribute showTime=\"true\" requires time zone support to be enabled when client and server are on different time zones.
+htmlformentry.error.noClientTimezone                    = The client time zone is not part of the submitted parameters and it is therefore not possible to determine the client time zone.
+
+htmlformentry.conditionui.historyOf.label               = History Of
+htmlformentry.conditionui.condition.required            = Condition required
+htmlformentry.conditionui.status.required               = Condition status required
+htmlformentry.conditionui.endDate.before.onsetDate.error= End date cannot be before onset date
+htmlformentry.conditionui.status.label                  = Status
+htmlformentry.conditionui.onsetdate.label               = Onset Date
```
dkayiwa, line 168: Did you add these spaces by accident?

api/src/main/resources/messages.properties
```diff
@@ -157,6 +157,9 @@ htmlformentry.chooseAnEncounterRole=Choose an Encounter Role
 htmlformentry.unknownProviderName=Unknown
 htmlformentry.providerPlaceHolder=Type identifier or name
 
+htmlformentry.error.handleTimezones                     = The attribute showTime=\"true\" requires time zone support to be enabled when client and server are on different time zones.
```
dkayiwa, line 160: Are these spaces intentional?

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryConstants.java
```diff
@@ -36,9 +36,15 @@
 	
 	public static final String GP_RESTRICT_ENCOUNTER_LOCATION_TO_CURRENT_VISIT_LOCATION = "htmlformentry.restrictEncounterLocationToCurrentVisitLocation";
 	
+	public final static String GP_TIMEZONE_CONVERSIONS = "htmlformentry.timezone.conversions";
+
```
dkayiwa, line 40: Are these the same as in UiFrameworkConstants.java?

icrc-jfigueiredo: Yes, because this project doesen't depent form UiFramework i cannot use that one.

api/src/main/resources/messages.properties
icrc-jfigueiredo: Should be some auto-formatting, already remove them.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryConstants.java
dkayiwa: Shouldn't the name be the same?

PR #220, opened 2021-01-07
Drugorder tests

mseaton:
This adds additional unit tests (and some related fixes identified in the course of doing so) for the new drugOrder tag.


mogoodrich (approved):
LGTM, though admittedly I didn't dig into too much detail... if there's a particular part you like me to dig into, lmk...

PR #222, opened 2021-01-14
Add null guard for more useful error message

brandones:
Ran into an NPE when there was an obs grouping concept missing. This improves the error from an NPE to

```
Caused by: java.lang.IllegalArgumentException: Cannot find concept for value CIEL:164352 in conceptId attribute value. Parameters: {answerConceptIds=CIEL:1527,PIH:GUARDIAN,PIH:PARTNER OR SPOUSE,CIEL:1528,PIH:SIBLING,PIH:OTHER RELATIVE, conceptId=CIEL:164352, labelCode=pihcore.contactRelationship}
        at org.openmrs.module.htmlformentry.element.ObsSubmissionElement.<init>(ObsSubmissionElement.java:174)
```

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/ObsGroupComponent.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/ObsGroupComponent.java
@@ -113,8 +113,14 @@ public class ObsGroupComponent {
 			Set<Integer> obsGroupComponentMatchLog = new HashSet<Integer>();
 			
 			for (ObsGroupComponent obsGroupComponent : obsGroupComponents) {
-				boolean questionMatches = obsGroupComponent.getQuestion().getConceptId()
-				        .equals(obs.getConcept().getConceptId());
+				Concept groupComponentQuestion = obsGroupComponent.getQuestion();
+				if (groupComponentQuestion == null) {
+					// The correct error should be thrown with useful contextual information from ObsSubmissionElement:174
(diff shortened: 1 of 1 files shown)


cioan (approved):

mogoodrich:
Thanks @brandones ... looks good to me... I was going to ask if we wanted to actually just throw an exception right when groupComponentQuestion == null instead of waiting for it to be thrown further down the line, since it's unclear what supportingRank will do in the case when we call "continue"...  but I see that the exception at ObsSubmissionElement:174 will be thrown regardless of the results of supportingRank, it was just a matter than the NPE was happening first before it hit that line?

A comment like "this is an error case, but we just make this null-safe and skip without throwing a NPE exception, with the expectation that a more useful error will be thrown at ObsSubmissionElement:174" might be a bit clearer...

mseaton:
I haven't looked at this in detail, but my initial thought was that if something is misconfigured we should try to catch it as early as possible (likely in the tag handler) and throw a BadFormDesignException.  Is that what this is doing @brandones ?  Or is that valid?

mogoodrich:
It is happening early on during the initialization of the Obs Submission Element, so it must be that this ObsGroup stuff must happen before that is hit.  I didn't dig further into why that was.

brandones:
@mseaton Yes, it's a misconfiguration. Catching it early does sound like the better way about it. I can look into that.

PR #223, opened 2021-01-22
Tiny refactor

brandones:
Just made this code a little easier to understand. I was attempting to catch a bad concept error in the tag handler as suggested [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/222#issuecomment-761029126), but got stuck before figuring it out. I don't really understand how the ObsGroupTagHandler works.


mseaton (approved):
Looks ok to me

PR #226, opened 2021-02-09
Html 759 - 3

mseaton:
This is the next in a series of relatively large commits that have all contributed to the refactoring and reworking of the drug order tag in the 4.0.0 release of HFE.

Here are the features / changes this PR aims to implement:

1. Ensures that the tag respects the existing global property "drugOrder.requireDrug" (only allows leaving drug or drug non coded blank if this is false), and also adds additional validation to ensure concept is required and only one of drug and drugNonCoded can be supplied.

2. Adds the ability to configure any of the drug order properties that are Concepts (eg. concept, discontinueReason, orderReason, doseUnits, route, etc) via a single grouping concept (this can be either a question with answers or a set with members).  Provide mechanisms to configure attributes within this mechanism to drive display labels (via message codes, concept name tags, and concept name types), css classes, and order of the resulting options rendered in within a dropdown widget.

3. Related to 2, modify the DropDownWidget and Option classes to support the ability to organize Options within <optgroup> groups and to apply css classes to both optgroup and option elements.

4. Changes the labels for the buttons used when editing an order within the original encounter it was placed in to "Edit" and "Delete" to reflect the implemented behavior which voids and recreates an Order in this circumstance rather than stop/discontinue.  Adds in the ability to Delete an existing Discontinue Order in the encounter in which it was placed, in order to ensure appropriate data management is supported.

5. Related to 4, display a warning message letting users know that Edit and Delete will void previous orders when used within the current encounter.

6. Continued improvements and tweaks to the rendered drug order view, and improvements to usability.


api/src/main/java/org/openmrs/module/htmlformentry/widget/DropdownWidget.java
```diff
@@ -54,19 +56,43 @@ public String generateHtml(FormEntryContext context) {
 				sb.append(" size=").append("\"" + size.intValue() + "\"");
 			}
 			sb.append(">");
+			String currentOptionGroup = "";
 			for (int i = 0; i < getOptions().size(); ++i) {
 				Option option = getOptions().get(i);
+				
+				String optionGroup = (option.getGroupLabel() == null ? "" : option.getGroupLabel().trim());
```
mogoodrich, line 63: It is worth adding a comment or refactoring this a bit to make it clear that if "optionGroup" is empty, this whole block of code doesn't do anything?  Was just trying to quickly confirm that for existing widgets this code does't do anything, and it took me a minute or so to follow how that works...

mogoodrich (approved):
Took a look, but, yeah, honestly, I'm going to trust your judgement on this.  Code looks well organized and well tested.    The new OptionGroup thing definitely could be useful outside of Drug Orders.  If it's usable out of the box with regular obs we should document in the HFE reference and point Ellen to it.

PR #228, opened 2021-02-24
HTML-771

mseaton:
Obs submission throws NPE when using radio style in combination with conceptIds


mogoodrich (approved):
LGTM, though seems so strange we had never run into this before...

PR #231, opened 2021-03-04
HTML-759 Rewrite drugOrder tag for OpenMRS 2.x

mseaton:
This PR is a significant refactor that aims to build on the existing work and provide a foundation to use the underlying logic in the drugOrder tag for all sorts of orders.  The changes are mostly non-functional - renaming classes and text from drug order specific nomenclature to order more broadly.

This PR also contains a couple of bug fixes and performance improvements.


mogoodrich (approved):
Yeah, let's just merge and test, I'm not going to be able to review this in much detail... :)

PR #234, opened 2021-03-08
Replace NPE in RadioButtonsWidget with RuntimeException

brandones:
Previously, when an `<obs>` tag was created for a Coded concept that has no answers in its answer set, and no answers were provided in the XML, an NPE would be thrown. This replaces that NPE with a RuntimeException explaining what went wrong.


brandones:
https://github.com/PIH/openmrs-config-pihemr/pull/140

mogoodrich (approved):
LGTM, merging, thanks @brandones !

PR #236, opened 2021-04-14
Html 773

mseaton:
HTML-773
Obs tag using answerDrugs="true" should have option to exclude retired drugs


mogoodrich (approved):
LGTM, and nice to clean up that that old 1.6/1.7 code as well!

PR #238, opened 2021-04-28
HTML-755: <encounterDate showTime="true"/> to support time zones.

mogoodrich:
#### Ticket:
https://issues.openmrs.org/browse/HTML-755

#### What I have done:
See below [here](https://github.com/openmrs/openmrs-module-htmlformentry/pull/238#issuecomment-829256150).

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java
@@ -1,17 +1,16 @@
 package org.openmrs.module.htmlformentry.widget;
 
-import java.text.SimpleDateFormat;
-import java.util.Calendar;
-import java.util.Date;
-
-import javax.servlet.http.HttpServletRequest;
-
 import org.openmrs.api.context.Context;
 import org.openmrs.module.htmlformentry.FormEntryContext;
 import org.openmrs.module.htmlformentry.FormEntryContext.Mode;
 import org.openmrs.module.htmlformentry.HtmlFormEntryConstants;
 import org.openmrs.module.htmlformentry.HtmlFormEntryUtil;
 import org.springframework.util.StringUtils;
+
+import javax.servlet.http.HttpServletRequest;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
 
 /**
  * A widget that allows the selection of a certain time-of-day. To handle both a date and time, see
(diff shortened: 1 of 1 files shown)


dkayiwa:
@mogoodrich did you see the failing test?

mogoodrich:
Nope, thanks @dkayiwa  I will check it out!

mogoodrich:
Yeah @dkayiwa that's a real error, my change was mistaken... @mks-d  @icrc-jfigueiredo  thoughts on how this is supposed to work?  It seems like the TimeWidget.generateEditMode needs to support both initialization via an initial value and via valAsCal depending on whether it's used as part of the DateTimeWidget or the ZonedDateTimeWidget.

Also, was it intentional to make the ZonedDateTimeWidget the widget that is always used with the EncounterDetailsSubmissionElement?  I thought this was something that might be controlled by the GP?

icrc-jfigueiredo:
> tentional to make the ZonedDateTimeWidget the widget that is always used with the EncounterDetailsSubmissionElement? I thought this was something that might be controlled by the G

Hello @mogoodrich , what do you mean by DateTimeWidget , because for encounterDate, if it have a time component (showTime=true), it is always ZonedDateTimeWidget.

mogoodrich:
@icrc-jfigueiredo I'll take a look at the error and the test again.

What I meant was (and I could be forgetting what we discussed) was that I thinking if the imezone.conversions GP was set to to false, it would *not* use the ZonedDateTime widget but just stick with the legacy DateTime widget.  @mks-d maybe can refresh my memory?

(The backstory here is that at PIH I didn't think we had any forms that used "showTime='true'" but it turns out that there is one that does... the good news it was never put into production and I don't think it ever will be, so we can work around this, but just brought these issues to light and was throwing stack traces when our smoke tests were running).

mogoodrich:
Just saw @mks-d 's question on the ticket, which is a good one.  A little more detail.

The form I'm referencing has this line, and after the changes in HTML-755, it stack traces when trying to open the form:

https://github.com/PIH/openmrs-config-pihemr/blob/master/configuration/pih/htmlforms/deathCertificate.xml#L457

This is because it hits this line and you basically get a NPE because initialValue is null:

https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java#L49

Using the old DateTimeWidget, initial value was set before the generateEditModeHtml was ever called:

https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/widget/DateTimeWidget.java#L74

However, the new ZonedDateTimeWidget never appears to set the initial value, but rather sets a valAsCal based on the initial value, and then that new parameter is passed into the generateEditModeHtml, which used it to set the initial value, except in the "hidden" block:

https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/widget/TimeWidget.java#L45

... which still uses the old means.

So I changed this to use valAsCal as well, but as @dkayiwa correctly points out breaks another test, and I suspect would break using the DateTimeWidget as-is.

So I'm not sure exactly what is going on, but it's difficult for me to follow the theory/plan behind the changes and something smells a bit... sorry if I'm missing it?

dkayiwa:
> if the imezone.conversions GP was set to to false, it would _not_ use the ZonedDateTime widget but just stick with the legacy DateTime widget

Yes this is exactly what i expected. I have some review comments along those lines on this pull request: https://github.com/openmrs/openmrs-module-htmlformentryui/pull/40#discussion_r621049524

mks-d:
@mogoodrich closing this as it was fixed here: https://github.com/openmrs/openmrs-module-htmlformentry/commit/84127687313ce5c7f8c25566c65447058b7ccf5b#diff-659683b728884be9ea909f2107ea9511a277f2d05514b03a4ee8a4698f3a38bdL50-R55

PR #243, opened 2021-06-28
HTML-561 Obs group support for Signs and Symptoms construct

brandones:
https://issues.openmrs.org/browse/HTML-561

Using

```xml
<obsgroup groupingConceptId="CIEL:1727" contextObs="CIEL:1728=CIEL:151">
  <obs conceptId="CIEL:1729" answerConceptId="CIEL:1065" style="checkbox" answerLabel="Douleur abdominale"/>
</obsgroup>
```

We can achieve the following behavior:

contextobs

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
@@ -534,8 +534,8 @@ public class FormEntryContext {
 	
 	/**
 	 * Finds whether there is existing obs created for the question concept and numeric answer, returns
-	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox" // TODO: that
-	 * can't be right
+	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox"
+	 * // TODO: the above comment does not make sense
 	 *
 	 * @param question - the concept associated with the Obs to acquire
 	 * @param numericAns - numeric answer given with <obs/> declaration
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
@@ -534,7 +534,8 @@ public class FormEntryContext {
 	
 	/**
 	 * Finds whether there is existing obs created for the question concept and numeric answer, returns
-	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox"
+	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox" // TODO: that
+	 * can't be right
 	 *
 	 * @param question - the concept associated with the Obs to acquire
 	 * @param numericAns - numeric answer given with <obs/> declaration
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
@@ -535,7 +535,7 @@ public class FormEntryContext {
 	/**
 	 * Finds whether there is existing obs created for the question concept and numeric answer, returns
 	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox"
-	 * // TODO: the above comment does not make sense
+	 * // TODO: the above comment does not make sense???
 	 *
 	 * @param question - the concept associated with the Obs to acquire
 	 * @param numericAns - numeric answer given with <obs/> declaration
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/action/ObsGroupAction.java
```diff
@@ -74,6 +76,31 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
...
+					}
+					members.removeAll(session.getSubmissionActions().getObsToVoid());
+					Obs existingContextObs = null;
+					for (Obs member : members) {
+						if (member.getConcept() == contextObsConcept && member.getValueCoded() == contextObsValue) {
+							existingContextObs = member;
+							members.remove(existingContextObs);
+							break;
+						}
+					}
```
mogoodrich, line 96: Can you explain what the above logic is doing?  I'm a bit confused by it.

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -1382,7 +1382,7 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
 		if (commentFieldWidget != null)
 			comment = commentFieldWidget.getValue(session.getContext(), submission);
 
-		// note that the rare use case where showCommentField=true and style=location won't work, as "org.openmrs.Location"
+		// note that style=location cannot be used with showCommentField=true, since "org.openmrs.Location"
```
mogoodrich, line 1385: Thanks, that's clearer, but shouldn't it be "... since 'org.openmrs.Location' is set in the comment field"

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -1403,12 +1403,13 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
 
 			// call this regardless of whether the new value is null -- the
 			// modifyObs method is smart
-			if (concepts != null)
+			if (concepts != null) {
 				session.getSubmissionActions().modifyObs(existingObs, concept, answerConcept, obsDatetime,
 				    accessionNumberValue, comment);
-			else
+			} else {
```
mogoodrich, line 1409: ugh, still some bracketless if/thens around, thanks for fixing!

mogoodrich (commented):
Generally looks good, thanks @brandones ! I couldn't quite follow the core code that does the submission, so I asked a question about it... but the test cases seem to have good coverage of all we need to handle.

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
brandones: `since "org.openmrs.Location" will override any user-entered comment` as you wrote it, seems good to me. We'd both changed this comment and when I merged master I just resolved the conflict in my favor :shrug:

api/src/main/java/org/openmrs/module/htmlformentry/action/ObsGroupAction.java
brandones: I added an explanatory comment. Does it make sense?

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
```diff
@@ -535,6 +535,7 @@ public Obs removeExistingObs(List<Concept> questions, Concept answer) {
 	/**
 	 * Finds whether there is existing obs created for the question concept and numeric answer, returns
 	 * that <obs> if any, Use this only when datatype is numeric and style="checkbox"
+	 * // TODO: the above comment does not make sense
```
mogoodrich, line 538: you are probably right, but let's add a question mark to this comment until we figure it out?

mogoodrich (approved):
Great, LMK, with one very small comment.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
brandones:
```suggestion
	 * // TODO: the above comment does not make sense???
```

brandones:
Updated the documentation: https://wiki.openmrs.org/display/docs/HTML+Form+Entry+Module+Reference#HTMLFormEntryModuleReference-%3Cobsgroup%3E

mogoodrich:
Sweet, thanks!

PR #249, opened 2021-11-03
Html 792

mseaton:
Addresses the following 2 issues:

1. Adds a validation to the OrderSubmissionElement that mimics the validation from the OrderService having to do with overlapping orders for the same orderable, and if an order is submitted that will fail back-end validation due to this, a new htmlform field validation error is displayed.

2. Fixes a bug with displaying validation issues for new orders in the order tag on an htmlform, which was introduced when the handling for new orders was adjusted to choose the orderable after choosing to place a new order.


cioan (approved):
Looks good @mseaton !

PR #254, opened 2021-11-19
HTML-794: Expand "allowFutureDates " to prevent selection of bad dates

mogoodrich:


mogoodrich:
So here's a first pass at this that will handle some use cases by restricting the date widget so that an invalid value shouldn't be enterable (server-side validation upon submit will still happen for use cases that are missed).

What this does:

Adds a "maxDate" parameter to the Date Widget
When a Date Widget is instantiated, it sets the max allowed value to the maxDate, if specified
When rendering a date widget for an obs of value date and "allowFutureDates" is false (or not specified) set the max date to:
Any existing encounter date, if present, else
Any pending encounter date, if present (I don't think there's actually a path where this will occur currently), else
Current date
BUT if there's a current obs value and it's after whatever was chosen out of the above three bullets, use that (so we don't inadvertently change an existing value)
What this doesn't do:

Since this is only set on form instantiation, it won't update the max date if the encounter date is changed; and note that this means that generally the max date will be set to "today" when entering a new form, because no encounter date is available when the form is initially rendered
We could potentially also add a change handler that listens for changes to the encounter date and updates the max date accordingly. (we'd still want to make sure we never change the currently selected date).

mogoodrich:
See comment on HTML-794 for further explanation:
https://issues.openmrs.org/browse/HTML-794

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -632,6 +633,7 @@ function setupDatePicker(jsDateFormat,yearsrange, jsLocale, displaySelector, val
 		gotoCurrent: true,
 		changeMonth: true,
 		changeYear: true,
+    maxDate: maxDate !== null ? jQuery.datepicker.parseDate('yy-mm-dd', maxDate) : null,
```
brandones, line 636:
```suggestion
		maxDate: maxDate !== null ? jQuery.datepicker.parseDate('yy-mm-dd', maxDate) : null,
```

brandones (approved):

mseaton (approved):
LGTM - Thanks @mogoodrich

PR #255, opened 2021-12-13
HTML-798 Add "allowPastDates" attribute to Obs tag

mseaton:
Adds a new option attribute to the obs tag for date obs that enables one to specify that past dates should be disallowed and lead to validation errors.  Note:  this is only enabled as server-side validation due to complexities in adjusting to the effective validation date when the encounter date field on the form changes.  If client-side validation is desired, that can be tackled in a separate ticket.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
@@ -110,6 +110,8 @@ public class ObsSubmissionElement implements HtmlGeneratorElement, FormSubmissio
 	
 	private boolean allowFutureDates = false;
 	
+	private boolean allowPastDates = true;
+	
 	private Concept answerConcept;
 	
 	private Drug answerDrug;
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -214,6 +216,9 @@ else if (conceptId == null && conceptIds == null)
 		if ("true".equalsIgnoreCase(parameters.get("allowFutureDates"))) {
 			allowFutureDates = true;
 		}
+		if ("false".equalsIgnoreCase(parameters.get("allowPastDates"))) {
+			allowPastDates = false;
+		}
```
mogoodrich, line 221: I'd be inclined to remove all the min date stuff if we aren't adding client-side validation as that is what the min and max date properties on Date Widget was intended for.  Thoughts?

mogoodrich (commented):
I added a comment on the code, not sure we want to add the "minDate" if we aren't actually using it when setting up the date picker with a min date value?  T

PR #257, opened 2022-01-06
HTML-805 - Form Submission Actions should not void obs prior to savin…

mseaton:
…g encounter

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
@@ -1,17 +1,6 @@
 package org.openmrs.module.htmlformentry;
 
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpSession;
-import java.io.StringWriter;
-import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
+import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.time.DateUtils;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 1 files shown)


mogoodrich (commented):
We should have unit test around this to confirm that an obs is properly voided (or maybe there's an existing one that already tests this?)

Otherwise, if this works, great!  Note that I did try some things similar to this (though not exactly in this order) and ran into issues where other tests started to fail... but if you are not seeing that, great... this generally seems like the good way to go about it...

mseaton:
My assumption is that we have a ton of unit test coverage in HFE, and there must be plenty existing tests around voiding obs, but yes I agree more unit testing would help if needed.  I also ran into failing tests _iff_ I tried to move the void block after the encounter save.  But this doesn't do that.  It leaves everything right where it was, it just eliminates the unnecessary explicit obsService.voidObs calls.

mogoodrich:
Sweet, thanks @mseaton .  I think we should just confirm that there's a test around a voiding an obs, and add one if needed (but I suspect you are right and they exist).

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -718,11 +719,16 @@ public void applyActions() throws BadFormDesignException {
 		
 		ObsService obsService = Context.getObsService();
 		
+		boolean patientUpdateRequired = submissionActions.getPatientUpdateRequired();
+		
 		if (submissionActions.getObsToVoid() != null) {
 			for (Obs o : submissionActions.getObsToVoid()) {
+				if (o.getEncounter() == null) {
```
dkayiwa, line 726: Is there any possibility of `o.getEncounter()` being `null` when `submissionActions.getPatientUpdateRequired()` is `false`?

mseaton: No idea.  This was just a sanity check to ensure that if obs are being edited outside of the context of an encounter that they are persisted later on by calling savePatient, since saveEncounter is never called in this case.

PR #259, opened 2022-01-29
HTML-808 Text Field Widget applies value class incorrectly on empty v…

mseaton:
…alue


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #273, opened 2022-08-05
HTML-818 - encounter.json controller throws an error

mseaton:


mogoodrich (approved):
Defer to you if this works, LGTM @mseaton

PR #274, opened 2022-09-28
HTML-820 - workflowState tag submits changes in separate transaction

mseaton:


api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -573,35 +577,6 @@ public void applyActions() throws BadFormDesignException {
...
-					} else if (outcomeObs.size() > 1) {
-						throw new IllegalStateException(
-						        "Unable to complete patient program as multiple outcome observations are recorded: "
-						                + outcomeObs);
-					}
-				}
-			}
-		}
-		
-		// TODO wrap this in a transaction
```
mseaton, line 604: The above was not removed, but moved and incorporated into the logic further below.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -667,54 +642,168 @@ public void applyActions() throws BadFormDesignException {
...
+			}
+			PatientProgram existing = HtmlFormEntryUtil.getCurrentOrNextFutureProgramEnrollment(toCreate.getPatient(),
+			    toCreate.getProgram(), toCreate.getDateEnrolled());
+			if (existing != null) {
+				// edit this enrollment to move its start date earlier
+				existing.setDateEnrolled(toCreate.getDateEnrolled());
+				programsToUpdate.add(existing);
+				ppCreateIter.remove();
+			}
+		}
```
mseaton, line 675: The above assumes that there is nothing stored on the programsToCreate that needs to be preserved aside from the enrollment date, as if it finds a matching program it will use it and the only thing it then updates on the existing program is the enrollment date.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -667,54 +642,168 @@ public void applyActions() throws BadFormDesignException {
...
+				pp = HtmlFormEntryUtil.getPatientProgramByProgram(programsToComplete, program);
+			}
+			// If the PatientProgram is not already being actioned, update existing program or create new program
+			if (pp == null) {
+				
+				if (getContext().getMode() == Mode.EDIT) {
+					pp = HtmlFormEntryUtil.getPatientProgramByProgramOnDate(patient, program, encounterDate);
+				} else {
+					pp = HtmlFormEntryUtil.getCurrentOrNextFutureProgramEnrollment(patient, program, encounterDate);
+				}
```
mseaton, line 718: This basically says - if you are trying to create a program, try to instead edit an existing program if the patient has an enrollment.  If they have an active enrollment on the encounter date, use it.  If this is an ENTRY and not an EDIT, then _also_ check if they have a future enrollment if they don't have an active one, and if they have a future enrollment, then adjust it's start back to this encounter date and use it. 

I don't entirely understand why this logic is in place, but implementing it in this way enabled me to get all of the unit tests to pass.  The EDIT mode condition, in particular, was needed to get , WorkflowStateTagTest.shouldNotEnrollInProgramOnEditIfPatientAlreadyEnrolledInProgram to pass.

api/src/main/java/org/openmrs/module/htmlformentry/FormSubmissionActions.java
```diff
@@ -630,30 +632,18 @@ public void completeProgram(Program program) {
...
-			patientProgram = HtmlFormEntryUtil.getPatientProgramByProgram(patientProgramsToCreate,
-			    state.getProgramWorkflow().getProgram());
-		}
-		
-		if (patientProgram == null) {
-			patientProgram = HtmlFormEntryUtil.getPatientProgramByProgram(patientProgramsToUpdate,
-			    state.getProgramWorkflow().getProgram());
 		}
-		HtmlFormEntryUtil.transitionToState(patient, patientProgram, state, encounter);
+		programWorkflowStatesToTransition.add(state);
```
mseaton, line 646: Removing the above code that actually transitions the state here was the main reason for this ticket.  This logic has been added to the FormEntrySession above.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -2547,110 +2601,6 @@ public static List<Provider> getAllProviders(boolean includeRetired) {
...
-		}
-		
-		patientProgram.getStates().add(newState);
-		
-		if (encounter.getEncounterDatetime().before(patientProgram.getDateEnrolled())) {
-			patientProgram.setDateEnrolled(encounter.getEncounterDatetime());
-		}
-		Context.getProgramWorkflowService().savePatientProgram(patientProgram);
-	}
-
```
mseaton, line 2653: The above logic was removed from this util class, and incorporated into the logic in the FormEntrySession.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -667,54 +642,168 @@ public void applyActions() throws BadFormDesignException {
...
+			
+			// If the state to transition into is not the same as the current state, transition into it
+			PatientState stateOnEncounterDate = HtmlFormEntryUtil.getPatientStateOnDate(pp, workflow, encounterDate);
+			if (stateOnEncounterDate == null || !stateOnEncounterDate.getState().equals(state)) {
+				
+				// Because htmlforms might update existing programs, and have retrospective entry, we do not
+				// simply call transitionToState here on the patientProgram.  Rather, we iterate over the states
+				// in the workflow, and try to insert the new state into the existing states where appropriate
+				
+				PatientState newState = new PatientState();
```
mseaton, line 742: All of the below logic was moved from the (previous) HtmlFormEntryUtil.transitionToState method that was removed below

mseaton (commented):
@mogoodrich I know it's been a long time, but would be great if you could review this with an eye towards how you originally wrote these tags and how you expect them to work, along with how the documentation says they should work.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
mogoodrich: We *could* decide we want to change this behavior, but there are are other tests in the class that explicitly test that a new program is created.

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
```diff
@@ -426,9 +427,9 @@ public void testResults(SubmissionResults results) {
...
 				
-				//Then: Workflow state Y is created with a start date of June 2011 and a stop date of Jan 2012.
-				// Workflow state X stays as is.
-				PatientProgram pp = assertProgram(results.getPatient(), TEST_PROGRAM, ONE_YEAR_AGO, null);
+				//Then: The initial state is ended, the middle state is added, and the end state stays as is
+				PatientProgram pp = assertProgram(results.getPatient(), TEST_PROGRAM, TWO_YEARS_AGO, null);
+				assertState(pp, START_STATE, TWO_YEARS_AGO, ONE_YEAR_AGO);
 				assertState(pp, MIDDLE_STATE, ONE_YEAR_AGO, TODAY);
 				assertState(pp, END_STATE, TODAY, null);
 			}
```
mseaton, line 435: Per @mogoodrich comments in the code review, this test was not doing the right thing.  This fixes things such that the initial enrollment starts 2 years ago, and this test demonstrates inserting a middle state between two existing states in an enrollment

api-tests/src/test/java/org/openmrs/module/htmlformentry/WorkflowStateTagTest.java
```diff
@@ -1155,14 +1156,14 @@ public void testEditedResults(SubmissionResults results) {
...
 				
-				// assert that the program has only one state
-				Assert.assertEquals(1, nonVoidedStates(pp));
-				
-				// assert that the start state of the state is correct
-				assertState(pp, START_STATE, ONE_YEAR_AGO, null);
+				PatientProgram pp2 = assertProgram(results.getPatient(), TEST_PROGRAM, TODAY, null);
+				Assert.assertNotNull(pp2);
+				Assert.assertNotEquals(pp1, pp2);
 			}
```
mseaton, line 1167: This test was originally passing, but upon review we found that it should not have been.  We fixed the code to remove the problematic behavior (creating 2 overlapping enrollments), and then updated this to test that it is done properly

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -667,8 +667,10 @@ public void applyActions() throws BadFormDesignException {
 			PatientProgram existing = HtmlFormEntryUtil.getCurrentOrNextFutureProgramEnrollment(toCreate.getPatient(),
 			    toCreate.getProgram(), toCreate.getDateEnrolled());
 			if (existing != null) {
-				// edit this enrollment to move its start date earlier
-				existing.setDateEnrolled(toCreate.getDateEnrolled());
+				// Shift the existing enrollment date earlier if needed to match the new enrollment date
+				if (toCreate.getDateEnrolled().before(existing.getDateEnrolled())) {
+					existing.setDateEnrolled(toCreate.getDateEnrolled());
+				}
```
mseaton, line 673: Per @mogoodrich comments, we only want to shift the enrollment date earlier, not later

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -696,99 +698,174 @@ public void applyActions() throws BadFormDesignException {
...
 			Date encounterDate = encounter.getEncounterDatetime();
 			
-			// First determine if the associated PatientProgram is already set to be created, updated, or completed
-			PatientProgram pp = HtmlFormEntryUtil.getPatientProgramByProgram(programsToCreate, program);
-			if (pp == null) {
-				pp = HtmlFormEntryUtil.getPatientProgramByProgram(programsToUpdate, program);
-			}
-			if (pp == null) {
-				pp = HtmlFormEntryUtil.getPatientProgramByProgram(programsToComplete, program);
+			// Try to determine if this is intended as an edit to an existing state
```
mseaton, line 709: This basically adds in the logic that was in the WorkflowStateSubmissionElement, so now all of the logic that determines what patientprogram changes and state transitions are needed for a given PatientState entry are handled in the form entry session.

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -696,99 +698,174 @@ public void applyActions() throws BadFormDesignException {
...
+			// by checking if a state in this workflow is active on the previous encounter date
+			PatientState stateToEdit = null;
+			if (Mode.EDIT.equals(context.getMode()) && previousEncounterDate != null) {
+				stateToEdit = HtmlFormEntryUtil.getPatientStateOnDate(patient, workflow, previousEncounterDate);
+				if (stateToEdit != null) {
+					pp = stateToEdit.getPatientProgram();
+					if (!programsToUpdate.contains(pp)) {
+						programsToUpdate.add(pp);
+					}
+				}
```
mseaton, line 719: This simplifies what was previously in the submission element - basically, if we identify that we are trying to edit a previously entered state, then update that state.  Otherwise create a new state

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -696,99 +698,174 @@ public void applyActions() throws BadFormDesignException {
...
+				if (pp == null) {
+					pp = HtmlFormEntryUtil.getPatientProgramByProgram(programsToComplete, program);
+				}
 				
-				for (PatientState currentState : pp.statesInWorkflow(workflow, false)) {
+				// If not already in the programs to create, update, or complete, see if there is already an active program
+				// We consider here both programs that are active on the encounter date and programs that start after
+				// the encounter date, as otherwise we'd have overlapping or abutting enrollments in the same program
+				if (pp == null) {
+					pp = HtmlFormEntryUtil.getPatientProgramByProgramOnDate(patient, program, encounterDate);
```
mseaton, line 790: Per @mogoodrich comments, this only matches on active enrollments, not future enrollments.  So this will not shift any existing enrollments to start earlier, but rather create a new enrollment (which could potentially be prior to another existing enrollment).

api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -696,99 +698,174 @@ public void applyActions() throws BadFormDesignException {
...
+						pp.setProgram(program);
+						pp.setDateEnrolled(encounterDate);
+						
+						// If there is another enrollment in the same program in the future, end this new enrollment at the start of that enrollment
+						PatientProgram nextEnrollment = HtmlFormEntryUtil.getCurrentOrNextFutureProgramEnrollment(patient,
+						    program, encounterDate);
+						if (nextEnrollment != null && nextEnrollment.getDateEnrolled().after(encounterDate)) {
+							pp.setDateCompleted(nextEnrollment.getDateEnrolled());
+						}
+
```
mseaton, line 805: The above was added per the above, in case a new enrollment is being added before an existing enrollment, in which case we need to end the new enrollment at the start of the existing enrollment

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -2283,6 +2283,17 @@ public static PatientState getPatientStateOnDate(PatientProgram pp, ProgramWorkf
...
+	/**
+	 * @return the state that is active in the given workflow, on the given date, for the given patient
+	 */
+	public static PatientState getPatientStateOnDate(Patient patient, ProgramWorkflow workflow, Date onDate) {
+		PatientProgram patientProgram = HtmlFormEntryUtil.getPatientProgramByWorkflow(patient, workflow);
+		if (patientProgram == null) {
+			return null;
+		}
+		return getPatientStateOnDate(patientProgram, workflow, onDate);
+	}
```
mseaton, line 2295: @mogoodrich per your comment added this here.

api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java
```diff
@@ -798,15 +798,15 @@ public void handleSubmission(FormEntrySession session, HttpServletRequest submis
...
-			else {
+			} else {
 				// we don't want to lose any time information just because we edited it with a form that only collects date,
 				// so we only update the date if the date has a time component or the actual date has changed
 				if (hasTimeComponent(date) || !stripTimeComponent(date).equals(stripTimeComponent(previousDate))) {
 					session.getContext().setPreviousEncounterDate(
 					    new Date(session.getSubmissionActions().getCurrentEncounter().getEncounterDatetime().getTime()));
 					session.getSubmissionActions().getCurrentEncounter().setEncounterDatetime(date);
+				} else {
+					session.getContext().setPreviousEncounterDate(previousDate);
```
mseaton, line 809: I uncovered a separate bug I believe, in which this previous encounter date was not getting set on the context in all circumstances.

api/src/main/java/org/openmrs/module/htmlformentry/element/WorkflowStateSubmissionElement.java
```diff
@@ -212,100 +191,8 @@ public Collection<FormSubmissionError> validateSubmission(FormEntryContext conte
...
-				PatientProgram patientProgram = HtmlFormEntryUtil.getPatientProgramByProgramOnDate(session.getPatient(),
-				    state.getProgramWorkflow().getProgram(), session.getEncounter().getEncounterDatetime());
-				
-				if (patientProgram != null) {
-					session.getSubmissionActions().getPatientProgramsToUpdate().add(patientProgram);
-				}
-				
-				session.getSubmissionActions().transitionToState(state);
-			}
+			ProgramWorkflowState state = Context.getProgramWorkflowService().getStateByUuid(stateUuid);
```
mseaton, line 194: All of the above logic was moved to the session

mseaton:
@mogoodrich not sure if you want to have another look over, but here are my changes based on your feedback and review.

PR #275, opened 2022-09-30
HTML-821, add to currentProgramWorkflowStatus support for inout UUID

cioan:


api/src/main/java/org/openmrs/module/htmlformentry/VelocityFunctions.java
```diff
@@ -355,6 +355,28 @@ public PatientState currentProgramWorkflowStatus(Integer programWorkflowId) {
...
+	}
+	
+	public PatientState currentProgramWorkflowStatus(String programWorkflowId, Date latestDate) {
+		ProgramWorkflow programWorkflow = HtmlFormEntryUtil.getWorkflow(programWorkflowId);
+		if (programWorkflow != null) {
+			return currentProgramWorkflowStatus(programWorkflow.getId(), latestDate);
+		}
+		return null;
+	}
+
```
mseaton, line 379: @cioan have you tested this in HFE yet, with a real form?  I suspect this will not work - velocity (AFAIK) is not fully type-safe, so has no real way to distinguish the method that takes a String from the one that takes an Integer.

The right solution here I believe (and I think this is what @mogoodrich was proposing) is to change the original method signature to take in Object.  Then, if this is an Integer, you will need to look up by id (current behavior), but if this is a String then you would use the HtmlformEntryUtil.getWorkflow(String) method.

Can you try changing to that please?

mogoodrich: The underlying HtmlFormEntry.getWorkflow(String) should handle both the Integer and String use case (it tries to parse the String as an Integer)... but @mseaton  I think you are saying that keeping both a currentProgramWorkflowStatus(String) and currentProgramWorkflowStatus(Integer) in Velocity functions will be problematic  (in which case I think we'd rename the latter method and have it take in a ProgramWorkflow, ie may it more a of a utiltiy method)?  @mseaton  @cioan

cioan: @mseaton , thank you for your comments. Yes, the change I made works fine with real forms. I actually tested with the Ob/GYN form that triggered this change. 
Also, it looks like the velocity functions are type-safe. There were already these two functions:
[public PatientState currentProgramWorkflowStatus(Integer programWorkflowId, String latestDateString)](https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/VelocityFunctions.java#L359)

[public PatientState currentProgramWorkflowStatus(Integer programWorkflowId, Date latestDate) {](https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/VelocityFunctions.java#L365)

which I think they have been working fine for a while. Should we change those too?

mseaton: @cioan - all good then, if this works.  I feel like it may not have in the past, but maybe that issue was resolved...

PR #276, opened 2023-01-03
HTML-822 - Order validation for overlapping drug orders should use en…

mseaton:
…counter date

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
@@ -169,12 +169,12 @@ public class OrderSubmissionElement implements HtmlGeneratorElement, FormSubmiss
 							handleRequiredField(ret, ctx, fs, "quantityUnits", newDrugOrder.getQuantityUnits());
 						}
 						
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
```diff
@@ -352,12 +352,16 @@ public boolean overlapsWithExistingDrugOrder(List<Order> existingOrders, DrugOrd
...
+	private boolean checkScheduleOverlap(FormEntryContext ctx, Order newOrder, Order existingOrder) {
+		Date newStart = newOrder.getEffectiveStartDate();
 		Date newStop = newOrder.getEffectiveStopDate();
 		Date existingStart = existingOrder.getEffectiveStartDate();
 		Date existingStop = existingOrder.getEffectiveStopDate();
 		
+		if (newStart == null) {
+			newStart = ctx.getBestApproximationOfEncounterDate();
+		}
+
```
mogoodrich, line 364: This is really the key change, correct (to use the best approximation method instead of "today" if no date explicitly assigned), and the rest is refactoring so that the "getBestApproximationOf..." method can be shared easiest?

mogoodrich (approved):
LGTM, thanks @mseaton , sending this back to you.

PR #277, opened 2023-01-03
HTML-823 - Add ability to store a provider from a list as an Obs

mseaton:


api-tests/src/test/java/org/openmrs/module/htmlformentry/RegressionTest.java
```diff
@@ -34,6 +37,7 @@ public class RegressionTest extends BaseHtmlFormEntryTest {
 	@Before
 	public void loadData() throws Exception {
 		executeVersionedDataSet("org/openmrs/module/htmlformentry/data/RegressionTest-data-openmrs-2.3.xml");
+		executeVersionedDataSet("org/openmrs/module/htmlformentry/data/providerRoles-dataset.xml");
```
mogoodrich, line 40: Seems kind of odd to apply this test data to this vast set of tests (which, admittedly, is a problem in itself for having grown so huge)

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -141,10 +142,6 @@
 	
 	private String answerSeparator = null;
 	
-	// these are for location and provider options
-	
-	private List<Option> locationOptions = new ArrayList<Option>();
```
mogoodrich, line 146: Oh, I see, you just moved this closer to where is was actually used?

mogoodrich (approved):
A few small questions, but LGTM @mseaton , thanks!

PR #278, opened 2023-01-03
HTML-824 - enrollinprogram tag needs to expand method for determining…

mseaton:
… encounter date

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
@@ -11,6 +11,7 @@ import org.openmrs.Location;
 import org.openmrs.Obs;
 import org.openmrs.Order;
 import org.openmrs.Patient;
+import org.openmrs.Visit;
 import org.openmrs.api.context.Context;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
```diff
@@ -936,9 +937,19 @@ public Date getBestApproximationOfEncounterDate() {
...
 		
-		// finally, fall back to current date if nothing else
 		if (encounterDateToTest == null) {
-			encounterDateToTest = new Date();
+			Date currentDate = (getDefaultEncounterDate() != null ? getDefaultEncounterDate() : new Date());
+			
+			// If there is a visit in the context and the current date is outside of this visit, use visit start date
+			if (getVisit() != null) {
+				Visit visit = (Visit) getVisit();
+				if (visit.getStopDatetime() != null && currentDate.after(visit.getStopDatetime())) {
```
mogoodrich, line 946: this is probably never possible, but should we also test that the encounter date is not before the visit start?

mogoodrich (approved):
LGTM.... one question, but you've probably thought it through better than me, so no need to change unless you think what I said makes any sense... :)

PR #279, opened 2023-01-09
HTML-824: enrollinprogram tag needs to expand method for determining …

mogoodrich:
…encounter date

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
@@ -930,29 +930,26 @@ public class FormEntryContext {
 	 */
 	public Date getBestApproximationOfEncounterDate() {
 		// if there's a pending update to the encounter date, use that
-		Date encounterDateToTest = getPendingEncounterDatetime();
+		Date encounterDatetime = getPendingEncounterDatetime();
 		
 		// if no pending, but there's an existing encounter, use that encounter date
-		if (encounterDateToTest == null && getExistingEncounter() != null) {
-			encounterDateToTest = getExistingEncounter().getEncounterDatetime();
+		if (encounterDatetime == null && getExistingEncounter() != null) {
+			encounterDatetime = getExistingEncounter().getEncounterDatetime();
 		}
 		
-		if (encounterDateToTest == null) {
+		if (encounterDatetime == null) {
 			Date currentDate = (getDefaultEncounterDate() != null ? getDefaultEncounterDate() : new Date());
-			
 			// If there is a visit in the context and the current date is outside of this visit, use visit start date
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
```diff
@@ -930,29 +930,26 @@ public void registerPropertyAccessorInfo(String property, String widgetId, Strin
 	 */
 	public Date getBestApproximationOfEncounterDate() {
 		// if there's a pending update to the encounter date, use that
-		Date encounterDateToTest = getPendingEncounterDatetime();
+		Date encounterDatetime = getPendingEncounterDatetime();
```
mogoodrich, line 933: tangential change... the variable name was annoying me...

mseaton: Changing the variable name is fine, I agree

api/src/main/java/org/openmrs/module/htmlformentry/FormEntryContext.java
```diff
@@ -930,29 +930,26 @@ public void registerPropertyAccessorInfo(String property, String widgetId, Strin
...
 			// If there is a visit in the context and the current date is outside of this visit, use visit start date
-			if (getVisit() != null) {
-				Visit visit = (Visit) getVisit();
-				if (visit.getStopDatetime() != null && currentDate.after(visit.getStopDatetime())) {
-					encounterDateToTest = visit.getStartDatetime();
-				}
+			if (getVisit() != null && ((Visit) getVisit()).getStopDatetime() != null
+			        && currentDate.after(((Visit) getVisit()).getStopDatetime())) {
+				encounterDatetime = ((Visit) visit).getStartDatetime();
 			} else {
```
mseaton, line 946: So, I actually can't see any meaningful difference in this code?  What exactly did you change?  Stylistically, I don't think this is an improvement, I prefer the code that was in place before.  What am I missing?

mogoodrich: The issue is when line 944 evaluates to true and 946 evaluates to false, line 951 never gets hit, and null gets returned.   Alternatively to my change, we could add another else at 948 that also sets encounterDate = currentDate.

Or we could rework the whole thing just to return at each point when it finds a match, but I didn't to have to retest all use cases after doing that.

mseaton: Oh, I see.  Yeah, I would have just removed the else block and just had a new "if encounter date == null, set to xxx" at the end.  But this is fine.

mseaton (approved):

PR #286, opened 2023-03-16
Age date functions

mseaton:


mseaton:
Yeah, I had other functions but I removed them since I found it confusing.  There are server-side encounter dates, but also could be encounter dates on the client side that are not yet persisted.

PR #288, opened 2023-03-28
HTML-828 Fix conditions tag edit mode to properly void and recreate t…

mseaton:
…he condition.

## Issue I worked on
see https://issues.openmrs.org/browse/HTML-828

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/ConditionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/ConditionElement.java
@@ -1,21 +1,10 @@
 package org.openmrs.module.htmlformentry.element;
 
-import static org.openmrs.module.htmlformentry.HtmlFormEntryConstants.FORM_NAMESPACE;
-import static org.openmrs.module.htmlformentry.HtmlFormEntryUtil.getControlId;
-import static org.openmrs.module.htmlformentry.HtmlFormEntryUtil.isEmpty;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/ConditionElement.java
```diff
@@ -68,8 +69,6 @@ public class ConditionElement implements HtmlGeneratorElement, FormSubmissionCon
 	@Override
 	public void handleSubmission(FormEntrySession session, HttpServletRequest submission) {
 		FormEntryContext context = session.getContext();
```
mseaton, line 71: The crux of this issue here is that the existing code is changing the state of the Condition objects that are attached to the Hibernate Session in the handleSubmission method of the ConditionElement, which is invoked prior-to and outside-the-transaction-of the overall form submission handling that occurs in the FormEntrySession.  So we need to save any manipulation that can affect persisted data until we are in the FormEntrySession applyActions method.  This is a common bug we see in HFE that needs to be addressed in this way.  The solution is to add instructions for how to change the Conditions into the FormSubmissionActions, and then to apply these during the transactional submission.

mogoodrich (approved):
LGTM @mseaton ... I didn't dig in super deeply, but 100% agree about the need move the transactional logic into submission actions.

And looks like you've added good test coverage, thanks!

PR #292, opened 2023-05-15
HTML-832 - Correctly pass toggle attribute into checkbox for valueDrug.

mseaton:
see https://issues.openmrs.org/browse/HTML-832


api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -876,7 +876,8 @@ else if (parameters.get("answerDrugIds") != null) {
 								throw new IllegalArgumentException("Cannot find Drug for answerDrugId: " + answerDrugId
 								        + " in answerDrugId attribute value. Parameters: " + parameters);
 							}
-							valueWidget = createCheckboxWidget(drug.getName(), "Drug:" + drug.getId().toString(), null);
```
mogoodrich, line 879: ah, so this was the place it was missing, I assume?

mogoodrich (commented):
LGTM, thanks @mseaton !

PR #293, opened 2023-07-13
HTML-833 - Allow obs datetime values to have a future time if on the …

mseaton:
…same day


mogoodrich (approved):
LGTM, thanks @mseaton

PR #294, opened 2023-08-18
HTML-834, reset order dateActivated to match encounter datetime

cioan:
@mseaton , the following change resolves the issue we had with the mental health form in PIH EMR. Please let me know if this looks OK to you? Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
@@ -43,6 +43,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 import java.io.StringWriter;
 import java.lang.reflect.Field;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java
```diff
@@ -573,6 +574,14 @@ public void applyActions() throws BadFormDesignException {
 					}
 					order.setPreviousOrder(previousOrder);
 					order.setEncounter(encounter);
+					if (encounter.getEncounterDatetime().after(order.getDateActivated())) {
+						// HTML-834, the encounterDate time got reset to current time but the order.dateActicated was set at midnight
+						if (DateUtils.isSameDay(encounter.getEncounterDatetime(), order.getDateActivated())
+						        && "00:00:00".equals(new SimpleDateFormat("HH:mm:ss").format(order.getDateActivated()))) {
```
mseaton, line 580: For checking if the date is "midnight", you should use the existing `!HtmlFormEntryUtil.hasTimeComponent(order.getDateActivated())`

mseaton (approved):
Looks good, just one comment to use an existing function to check about the date midnight rather than use the simple date format check.

cioan:
Thanks @mseaton ! I have updated the PR.

PR #297, opened 2023-09-26
HTML-836 - Drugs should be sorted alphabetically if an explicit order is not set

mseaton:
## Issue I worked on
see https://issues.openmrs.org/browse/HTML-836


mogoodrich (approved):
LGTM, thanks @mseaton

PR #304, opened 2024-08-13
HTML-844 - Encounter fails to save if entered at the same time as vis…

mseaton:
…it but hideSeconds is set to true

See:  https://openmrs.atlassian.net/browse/HTML-844


mogoodrich (approved):
LGTM, thanks @mseaton

PR #312, opened 2025-04-17
HTML-855 - Add utility and service methods to parse an xml file into …

mseaton:
…form, htmlform, and form resources

see https://issues.openmrs.org/browse/HTML-855


ibacher (approved):
LGTM

PR #314, opened 2025-04-30
HTML-858 - Unescape submitted xml on htmlform admin page.

mseaton:
see https://issues.openmrs.org/browse/HTML-858

### DIFF
--- a/omod/src/main/java/org/openmrs/module/htmlformentry/web/controller/HtmlFormController.java
+++ b/omod/src/main/java/org/openmrs/module/htmlformentry/web/controller/HtmlFormController.java
@@ -4,6 +4,7 @@ import javax.servlet.http.HttpSession;
 import java.text.SimpleDateFormat;
 import java.util.List;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


dkayiwa:
@mseaton our practice has been to unescape only when we are going to use/parse the xml. But on storage in the database, we still wanted it escaped. Otherwise it would reintroduce the the Stored XSS attacks that were reported during pentesting.

mogoodrich:
I'll defer to @dkayiwa  on this one...

mseaton:
> our practice has been to unescape only when we are going to use/parse the xml. But on storage in the database, we still wanted it escaped. Otherwise it would reintroduce the Stored XSS attacks that were reported during pentesting

Hmm... I'm not really sure then.  Most forms are stored in the DB un-escaped at the moment, and are loaded that way via initializer.  And any code that aims to parse this xml and turn it into a form will need to unescape it and parse it.  This includes validating the form on save, rendering it into a schema, etc.

Is there a concrete example we can consider where an XSS attack could occur that we can look at with this?

Looping in @ibacher as well...

ibacher:
This is a tricky one. HFE is kind of, by its very nature, vulnerable to XSS, but simply encoding it in the database won't really address that, since the form would need to be decoded before its rendered. I think it's probably ok to store the unescaped XML in this instance.

mseaton:
Thanks @ibacher and @dkayiwa .  I've merged this in to get the team unblocked that rely on this.  Generally though, we should be trying to sunset this.

PR #316, opened 2025-05-15
HTML-861 - Support creation of all test order fields supported in Ope…

mseaton:
…nMRS 2.5.0+

See:  https://openmrs.atlassian.net/browse/HTML-861

### DIFF
--- a/api-tests/pom.xml
+++ b/api-tests/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.openmrs.module</groupId>
 		<artifactId>htmlformentry</artifactId>
-		<version>5.5.0-SNAPSHOT</version>
+		<version>6.0.0-SNAPSHOT</version>
 	</parent>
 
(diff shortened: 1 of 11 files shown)


mseaton:
> LGTM!
> 
> I hadn't been following the discussion fully, but is there any merit in changing this all directly to Service Order in instead of Test Order?

It's a good point.  I had tried that initially and got some error due to (I think) the fact that ServiceOrder is abstract, but those may be mitigated now.  I'll try switching TestOrder to ServiceOrder and see if things still build and work.

mseaton:
> > LGTM!
> > I hadn't been following the discussion fully, but is there any merit in changing this all directly to Service Order in instead of Test Order?
> 
> It's a good point. I had tried that initially and got some error due to (I think) the fact that ServiceOrder is abstract, but those may be mitigated now. I'll try switching TestOrder to ServiceOrder and see if things still build and work.

OK, that worked @mogoodrich .  Now supports any service order.  Going to merge, let me know if you have any followup thoughts.

PR #322, opened 2025-08-27
HTML-868 - Order widget should make drug ingredients available

mseaton:
see https://issues.openmrs.org/browse/HTML-868


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #329, opened 2026-03-05
HTML-760: Fix relationship tag transaction bug

arti-nova:
HTML-760: Fix relationship tag transaction bug

JIRA: https://openmrs.atlassian.net/browse/HTML-760

Summary:
This PR fixes the relationship tag transaction bug where relationships were being persisted outside the form submission transaction. Previously, relationships created using the relationship tag were saved immediately, resulting in separate synchronization records.

Now:
Relationship creation is queued in submission actions and processed within the same transaction as form submission, ensuring transactional consistency.

Changes:
- Updated RelationshipSubmissionElement to queue relationship creation instead of directly persisting relationships.
- Added integration test to verify relationship is not saved outside the submission transaction.

Testing:
- mvn clean install -DskipTests=false
- All tests passing.

Notes:
- No unrelated files were modified.
- Code follows module coding conventions.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/RelationshipSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/RelationshipSubmissionElement.java
@@ -368,9 +368,7 @@ public class RelationshipSubmissionElement implements HtmlGeneratorElement, Form
 							rel.setPersonA(relatedPerson);
 						}
 						rel.setRelationshipType(r);
-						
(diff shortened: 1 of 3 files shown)


krisha-9 (commented):
The change from directly saving the relationship to queuing it via submissionActions looks good and aligns better with the form submission workflow. 

One suggestion: consider adding validation or checks to ensure that the relationship object is not null and avoids unintended duplicates before adding it to the queue.

PR #330, opened 2026-03-06
HTML-878: Order Tag should support placing orders for the same drug a…

mogoodrich:
…t an earlier date if the expiry date doesn't overlap

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
@@ -8,6 +8,7 @@ import org.openmrs.CareSetting;
 import org.openmrs.Concept;
 import org.openmrs.Drug;
 import org.openmrs.DrugOrder;
+import org.openmrs.Duration;
 import org.openmrs.Encounter;
 import org.openmrs.FreeTextDosingInstructions;
 import org.openmrs.Order;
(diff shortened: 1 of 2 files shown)


mogoodrich (commented):
Thoughts @mseaton ?  This basically copies the logic getAutoExpireDate from the BaseDosingInstructions... (we can't call it directly since it depends on the effective start date being set)... though thinking about it now this assumes there is no custom dosing instructions set up?

api/src/main/java/org/openmrs/module/htmlformentry/element/OrderSubmissionElement.java
```diff
@@ -352,7 +355,7 @@ public boolean overlapsWithExistingDrugOrder(FormEntryContext ctx, List<Order> e
 		return false;
 	}
 	
-	private boolean checkScheduleOverlap(FormEntryContext ctx, Order newOrder, Order existingOrder) {
+	private boolean checkScheduleOverlap(FormEntryContext ctx, DrugOrder newOrder, Order existingOrder) {
```
mogoodrich: This wasn't being used outside the context of a drug order, so I changed the signature, but we could switch this back to an Order and I could do some testing/casting below to DrugOrder as needed if you'd prefer that.

mseaton (approved):
Seems fine to me @mogoodrich .  At the end of the day, it was never the intention within HFE to add stricter validation than is in place in core, it was just trying to add validation earlier in order to provide better information and messages to users.  So as long as the orders save successfully, no reason why HFE should stand in the way.

PR #331, opened 2026-05-14
HTML-881: add restrictToSessionVisitLocation parameter to restrict visit locations

cioan:
@mogoodrich , this was very similar to the restrictToCurrentVisitLocation.
Here is the lab results form after I added this new parameter:

Screenshot 2026-05-14 at 11 50 05 AM


mogoodrich (approved):
LGTM, thanks @cioan .  Let's make sure we add this attribute to the wikie as well:

https://openmrs.atlassian.net/wiki/spaces/docs/pages/25475242/HTML+Form+Entry+Module

cioan:
Thanks @mogoodrich ! I will add the new parameter to the wiki.

PR #332, opened 2026-05-14
HTML-880: Required fields should not be required if they are hidden v…

mogoodrich:
…ia a "controls/when/then" functionality


mseaton (approved):

cioan (commented):
Looks good! Thanks @mogoodrich !

PR #333, opened 2026-05-21
HTML-882: Create tag to schedule appointment

chibongho:
Adds a new `<ScheduleAppointment>` tag that allows for scheduling (Bahmni) appointments. Attributes documented in https://openmrs.atlassian.net/wiki/spaces/docs/pages/25476979/HTML+Form+Entry+Module+Reference#%3CscheduleAppointment%3E

Testing done:
- Added functional test
- Manuel test:
https://github.com/user-attachments/assets/f238c3ca-2e8f-427a-a703-c5a3abdbe5ad


cioan (approved):
Looks good! Thanks @chibongho !

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+import org.openmrs.module.htmlformentry.widget.DateWidget;
+import org.openmrs.module.htmlformentry.widget.DropdownWidget;
+import org.openmrs.module.htmlformentry.widget.TimeWidget;
+import org.openmrs.module.htmlformentry.widget.ErrorWidget;
+import org.openmrs.module.htmlformentry.widget.Option;
+import org.openmrs.module.htmlformentry.widget.TextFieldWidget;
+
+public class ScheduleAppointmentElement implements HtmlGeneratorElement, FormSubmissionControllerAction {
+
+	private static final String DEFAULT_LOCATION_TAG = "Appointment Location";
```
mogoodrich, line 45: Let's just default to all locations if no tag is specified?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+			return;
+		}
+
+		optional = "true".equalsIgnoreCase(parameters.get("optional"));
+		if (optional) {
+			scheduleChoiceWidget = new TextFieldWidget();
+			context.registerWidget(scheduleChoiceWidget);
+		}
+
+		// --- Location ---
```
mogoodrich, line 88: Can we add a "restrictToVisitLocation" atttribute, following the pattern of the same attribute for the encounterLocation tag (see the wiki for details)?

mogoodrich (changes requested):
Scary. Claude understands HFE.

@chibongho  looks good, one function change and one additional feature (so we will be able to use this in Haiti if needed), see my comments.  Otherwise, looks good.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+		locationWidget = new DropdownWidget();
+		locationErrorWidget = new ErrorWidget();
+		locationWidget.addOption(new Option("", Context.getMessageSourceService().getMessage(
```
mseaton, line 91: I've never seen this syntax used with the messageSourceService before.  It's not wrong per-se, and might even be better, but not consistent with other code I would suspect.  How is this done in other elements?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+		timeWidget.setHideSeconds(true);
+		context.registerWidget(dateWidget);
+		context.registerWidget(timeWidget);
+		startDateTimeWidget = new DateTimeWidget(dateWidget, timeWidget);
+		startDateTimeErrorWidget = new ErrorWidget();
+		context.registerWidget(startDateTimeWidget);
+		context.registerErrorWidget(startDateTimeWidget, startDateTimeErrorWidget);
+
+		// --- Provider ---
+		providerWidget = new DropdownWidget();
```
mseaton: This has historically been a major source of slowness on forms, as some implementations have thousands of providers.  We should look at how other provider widgets are configured and either default to one that is an ajax-autocomplete and if dropdown lists are needed, allow that to be configured.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+		context.registerWidget(providerWidget);
+
+		// --- Note ---
+		noteWidget = new TextFieldWidget();
+		noteWidget.setTextArea(true);
+		context.registerWidget(noteWidget);
+	}
+
+	@Override
+	public String generateHtml(FormEntryContext context) {
```
mseaton, line 167: This all appears to hard code layout pretty significantly.  We likely want to - at the very least - add css classes on all elements and support and id on the top-level element - to allow the layout to be fully customized if needed via either CSS or Javascript.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+	private ErrorWidget typeErrorWidget;
+
+	private ErrorWidget startDateTimeErrorWidget;
+
+	// stored for duration lookup at submission time
+	private List<AppointmentServiceDefinition> services = new ArrayList<>();
+
+	public ScheduleAppointmentElement(FormEntryContext context, Map<String, String> parameters) {
+		if (context.getMode() != Mode.ENTER) {
+			return;
```
mseaton, line 78: So...what happens if you are editing a form or viewing a form?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,354 @@
...
+
+	private String fieldRow(String label, String widgetHtml) {
+		return "<p><label>" + label + "</label><span>" + widgetHtml + "</span></p>";
+	}
+
+	/**
+	 * Parses the optional {@code type} tag attribute into a list of allowed AppointmentKind values.
+	 * Accepts a comma-separated list, optionally wrapped in brackets, e.g. {@code [scheduled,walkIn]}.
+	 * Matching is case-insensitive. Returns all kinds when the attribute is absent or empty.
+	 */
```
mseaton: This seems very non-standard.  What's with supporting brackets?  This should just do whatever everything else in HFE already does.

api/src/main/java/org/openmrs/module/htmlformentry/handler/ScheduleAppointmentTagHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.openmrs.module.htmlformentry.FormEntrySession;
+import org.openmrs.module.htmlformentry.FormSubmissionController;
+import org.openmrs.module.htmlformentry.element.ScheduleAppointmentElement;
+
+public class ScheduleAppointmentTagHandler extends SubstitutionTagHandler {
+
+	@Override
+	protected List<AttributeDescriptor> createAttributeDescriptors() {
+		return Collections.unmodifiableList(new ArrayList<AttributeDescriptor>());
```
mseaton: I would expect to see something in here, not just an empty list?

api/src/main/java/org/openmrs/module/htmlformentry/appointment/ScheduleAppointmentAbstractor.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+
+import org.openmrs.api.context.Context;
+import org.openmrs.module.appointments.model.Appointment;
+import org.openmrs.module.appointments.service.AppointmentsService;
+
+/**
+ * Calls to the AppointmentService are abstracted here to avoid runtime class loading issues in
+ * instances where the Appointments module is not present.
+ */
```
mseaton: This class seems unnecessary.  Can we just add this method to the element and eliminate the need for this class?

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -116,6 +116,9 @@
                 <entry key="appointments">
                     <bean class="org.openmrs.module.htmlformentry.handler.AppointmentsTagHandler"/>
                 </entry>
+                <entry key="scheduleAppointment">
+                    <bean class="org.openmrs.module.htmlformentry.handler.ScheduleAppointmentTagHandler"/>
+                </entry>
```
mseaton, line 121: This probably only works here because the tag handler has no references at all to any appointment module classes.

mseaton (changes requested):
Generally looks like a good start - a few comments.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: Good point.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: Oh, right... @chibongho  didn't we wanted to display a message in the case of the "edit" saying that you needed to go to the appt page to edit any appointments?  I forget the messaging we came up with for view mode.

chibongho: We don't have a way to link the appointment to the creating encounter, so we just show the message "To view or manage appointments for this patient, use the Appointments app." (key `htmlformentry.scheduleAppointment.viewModeMessage`) in `generateHtml()`.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: The O3 appointments app hard-codes the tag to "Appointment Location" (see [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/691ba13eaf68a5540c91355a714f326da7efe324/packages/esm-appointments-app/src/form/appointments-form.workspace.tsx) and [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/04fba64d5f8a0889a007fba475b667a1ff3034a2/packages/esm-appointments-app/src/constants.ts)). Should we keep it for consistency?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll simplify it to just the single arg function `Context.getMessageSourceService().getMessage(messingId)`

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll change it to an Ajax autocomplete field instead, similar to `ProviderAjaxAutoCompleteWidget`.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll change it to `TagUtil.parseListParameter()`

api/src/main/java/org/openmrs/module/htmlformentry/handler/ScheduleAppointmentTagHandler.java
chibongho: This list only needs to contain attributes whose values might reference OpenMRS objects (like provider uuid). HtmlFormExporter uses this list to include the referenced objects for export. `<scheduleAppointment>` only has plain string attributes.

api/src/main/resources/moduleApplicationContext.xml
chibongho: This is following the same pattern as the `AppointmentsTagHandler`. Any suggestions?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll add a wrapper `<div>` with id.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: The `locationTag` attribute should cover this, right? Should we add it in addition to that?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: Fine with me!

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: This works for me.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: No, this is different, this means (on top of any tag-based restriction) to restrict to locations that are descendants of (or equal to) the current session location.

Use case would be a combined  multi-facility system like the new ZL system--at Cange, we only want the user to be prompted with the Cange locations that are tagged as Appointment locations, not all the Appointment locations in the system.

mogoodrich (commented):
@chibongho added a few comments in response to your comments!

api/src/main/resources/moduleApplicationContext.xml
mseaton: Yeah, looks like this will be ok since there are no references to the appointments module directly in this class...

mseaton (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/appointment/ScheduleAppointmentAbstractor.java
chibongho: The method is being called in [FormEntrySession.java](https://github.com/openmrs/openmrs-module-htmlformentry/blob/1db53156682439f71ae6c61016132df373cf34ab/api/src/main/java/org/openmrs/module/htmlformentry/FormEntrySession.java). I don't see elements being instantiated there, so I don't think we should put it in the element class. That said, I think I can move this into the existing [AppointmentAbstractor.java](https://github.com/openmrs/openmrs-module-htmlformentry/blob/aa389015f814eb6288a3017300043f0d2e116425/api/src/main/java/org/openmrs/module/htmlformentry/appointment/AppointmentsAbstractor) class.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mseaton: Is this being handled then?  I don't see this being addressed?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mseaton: HtmlformentryUtil.translate covers this FWIW

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mseaton: @mogoodrich is this consistent with how the javascript functions and existing widgets would expect id to be used?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: Yes, that message is shown in view / edit mode.

image
image

mseaton: Not sure I love this.  I would think we would want to store something that would allow us to link these together.  For example, we could store an Obs that has a reference to the uuid of the created appointment.

mogoodrich: I agree it's not great, but this is what we agreed upon as a design for a phase one... there was also a desire on Fiona's part to drive them to the new appointments UI as much as possible.

That being said, using an obs to reference the appointment uuid does seem a reasonable way to implement this.

mogoodrich (approved):

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: I *believe* but can you take a look and confirm this, @chibongho ?

chibongho: There are exceptions, but elements usually have top level wrapper with the id set similar to what we have here (see [this](https://github.com/openmrs/openmrs-module-htmlformentry/blob/a10d49333d89a1890eda1a636587e3af0a44ea93/api/src/main/java/org/openmrs/module/htmlformentry/element/RelationshipSubmissionElement.java) and [this](https://github.com/openmrs/openmrs-module-htmlformentry/blob/c9d9bd305613042614b2b1cd1f9e86ebbf240e73/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java)), while widgets usually have their ids set to field names ([here](https://github.com/openmrs/openmrs-module-htmlformentry/blob/444e0c599fd896fd24ccc71946b35ca24a6ca1f2/api/src/main/java/org/openmrs/module/htmlformentry/widget/ProviderAjaxAutoCompleteWidget.java) and [here](https://github.com/openmrs/openmrs-module-htmlformentry/blob/663d32733fe0109165b5633c7e59ae7112d62a41/api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java))

mogoodrich: Works for me!

chibongho:
Closing this after discussion. We will instead build a sub-form to record as observations the fields required for scheduling appointment.

PR #334, opened 2026-05-27
HTML-883: add personAttribute tag for viewing/entering/editing person attributes

mogoodrich:
See: https://openmrs.atlassian.net/browse/HTML-883


api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
```diff
@@ -0,0 +1,285 @@
...
+ * <p>When multiple non-voided attributes of the requested type exist for the patient, the most
+ * recently created one is used and a warning is logged.
+ */
+public class PersonAttributeSubmissionElement implements HtmlGeneratorElement, FormSubmissionControllerAction {
+
+	private static final String FORMAT_STRING = "java.lang.String";
+
+	private static final String FORMAT_CONCEPT = "org.openmrs.Concept";
+
+	private static final String FORMAT_LOCATION = "org.openmrs.Location";
```
mseaton: These constants seem unnecessary when you can and should just get the name off the class object itself.  If you want to keep the constants, assign them values based on `XXX.class.getName()`

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
```diff
@@ -0,0 +1,285 @@
...
+	private Widget valueWidget;
+
+	/** The error widget (rendered in ENTER / EDIT modes only). */
+	private ErrorWidget errorWidget;
+
+	/** The existing non-voided attribute (most recent if there are several), or null. */
+	private PersonAttribute existingAttribute;
+
+	/** Format of the attribute type (java.lang.String / org.openmrs.Concept / org.openmrs.Location). */
+	private final String format;
```
mseaton: Do you need to keep this here as something independent from PersonAttributeType?  Can't you just do `attributeType.getFormat()` if you need it?

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
```diff
@@ -0,0 +1,285 @@
...
+				        + "\"; using most recent.");
+				matching.sort(Comparator.comparing(PersonAttribute::getDateCreated).reversed());
+			}
+			existingAttribute = matching.isEmpty() ? null : matching.get(0);
+		}
+
+		// ---- 3. Build widget by format (all modes, including VIEW) --------------------
+		errorWidget = new ErrorWidget();
+
+		if (FORMAT_STRING.equals(format)) {
```
mseaton: Just use the XXX.class.getName() here instead of these constants

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
```diff
@@ -0,0 +1,285 @@
...
+			w.setInitialValue(existingAttribute.getValue());
+		}
+		return w;
+	}
+
+	private Widget buildConceptWidget(FormEntryContext context, Map<String, String> parameters)
+	        throws BadFormDesignException {
+
+		String answerConceptIds = parameters.get("answerConceptIds");
+		if (!StringUtils.hasText(answerConceptIds)) {
```
mseaton: I believe PersonAttributeType has a `foreignKey` property that may be intended to act as a pointer to the allowed value set.  eg. The foreignKey could point to a concept that is a question with answers (or possibly a set with members, not sure) and this should define the set of allowed answers.  We should likely support that here so that this handles the basics as expected.

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
```diff
@@ -0,0 +1,285 @@
...
+				existingAttribute.setValue(value);
+				existingAttribute.setChangedBy(Context.getAuthenticatedUser());
+				existingAttribute.setDateChanged(new Date());
+			}
+		} else {
+			if (StringUtils.hasText(value)) {
+				// Create a new attribute
+				PersonAttribute newAttribute = new PersonAttribute();
+				newAttribute.setAttributeType(attributeType);
+				newAttribute.setValue(value);
```
mseaton: I'm not 100% sure this is the right way to set the value, can you just double check?  Seems like this will set the primary key value of the concept or location if those are the types?  It would be better if we could call a method to set the value to the actual Concept or Location that we parse here, and then let core store that as a string value however it sees fit, to protect against any changes in the future, but if this is the way to do it, it's fine I guess

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
mogoodrich: Fair. Fixed.

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
mogoodrich: Fair.  Over-optimization by Claude I should have fixed.

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
mogoodrich: Oh, fair, I had forgotten about this. I will fix. Arguably, we shouldn't even allow answerConceptIds, but I will keep that since it's been created.

api/src/main/java/org/openmrs/module/htmlformentry/element/PersonAttributeSubmissionElement.java
mogoodrich: Good catch.  The "seralize" method below wasn't something I was familiar, but I double-checked and Claude is indeed correct.

Claude: The reviewer's concern is fully valid and there's a better API available.
  
  The key findings:
  - PersonAttribute.setValue(String) is the only setter ? no typed API
  - But both Concept and Location implement Attributable<T>, which has a serialize() method
  - Concept.serialize() ? "" + getConceptId(); Location.serialize() ? "" + getLocationId()
  - getHydratedObject() uses hydrate(getValue()) to reconstruct the typed object
  - The reviewer is right: we should parse the submitted value to the actual object and call serialize() on it, letting core determine the storage format
  
  Currently we pass the raw widget string directly. The fix is to look up the typed object and call serialize() ? exactly as Attributable intends.

mogoodrich (commented):
Updated per code review (and one other inconsistency I found).

fyi @mseaton  if you want to take another look.

I am going to test and then merge.

PR #336, opened 2026-06-08
HTML-884: Move view-mode html generation into server-side for order w…

mseaton:
…idget

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
@@ -28,8 +28,10 @@ import org.openmrs.util.OpenmrsUtil;
 import javax.servlet.http.HttpServletRequest;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
```diff
@@ -444,28 +453,268 @@ public String generateHtml(FormEntryContext context) {
...
+				}
+				if (order.getAction() == Order.Action.NEW) {
+					newOrders.add(order);
+				} else {
+					existingOrders.add(order);
+				}
+			}
+		}
+
+		appendOrderSection(html, "orderwidget-existing-order-section",
```
cioan, line 516: Would be possible to skip this section if empty/no orders?

api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
```diff
@@ -444,28 +453,268 @@ public String generateHtml(FormEntryContext context) {
...
+		    translate("htmlformentry.orders.existingOrdersViewTitle"),
+		    translate("htmlformentry.orders.noOrders"),
+		    existingOrders, allOrdersById, encDate, currentEncId, context);
+
+		appendOrderSection(html, "orderwidget-new-order-section",
```
cioan, line 521: Would be possible to skip this section if empty?

cioan (approved):
Looks good, @mseaton ! Just couple minor comments to skip the empty lines with no data. Thanks!

api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
mseaton: Good thoughts, let me work this up and see

api/src/main/java/org/openmrs/module/htmlformentry/widget/OrderWidget.java
mseaton: Ditto

PR #337, opened 2026-06-10
HTML-885: Add a way to input provider obs with autocomplete search

chibongho:
See: https://openmrs.atlassian.net/browse/HTML-885

The `ProviderAjaxAutoCompleteWidget` is already created, but only used for the `<EncounterProviderAndRole>` tag. This PR adds function to input provider as obs with an autocomplete search.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
@@ -282,6 +282,8 @@ public class HtmlFormEntryUtil {
 			} else if (value instanceof Person) {
 				Person person = (Person) value;
 				obs.setValueText(person.getId().toString() + " - " + person.getPersonName().toString());
+			} else if (value instanceof Provider) {
+				obs.setValueText(((Provider) value).getId().toString());
 			} else {
 				obs.setValueText(value.toString());
 			}
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
@@ -30,8 +30,10 @@ import org.openmrs.module.htmlformentry.widget.CheckboxWidget;
 import org.openmrs.module.htmlformentry.widget.ConceptSearchAutocompleteWidget;
 import org.openmrs.module.htmlformentry.widget.DateTimeWidget;
 import org.openmrs.module.htmlformentry.widget.DateWidget;
+import org.openmrs.module.htmlformentry.util.MatchMode;
 import org.openmrs.module.htmlformentry.widget.DropdownWidget;
 import org.openmrs.module.htmlformentry.widget.DynamicAutocompleteWidget;
+import org.openmrs.module.htmlformentry.widget.ProviderAjaxAutoCompleteWidget;
 import org.openmrs.module.htmlformentry.widget.ErrorWidget;
 import org.openmrs.module.htmlformentry.widget.NumberFieldWidget;
 import org.openmrs.module.htmlformentry.widget.Option;
@@ -241,7 +243,7 @@ public class ObsSubmissionElement<T extends FormEntryContext> implements HtmlGen
 		isLocationObs = "location".equals(parameters.get("style")) || "location_radio".equals(parameters.get("style"))
 		        || "location_dropdown".equals(parameters.get("style"));
 		isProviderObs = "provider".equals(parameters.get("style")) || "provider_radio".equals(parameters.get("style"))
-		        || "provider_dropdown".equals(parameters.get("style"));
+		        || "provider_dropdown".equals(parameters.get("style")) || "provider_autocomplete".equals(parameters.get("style"));
 		isRadioSet = "radio".equals(parameters.get("style")) || "location_radio".equals(parameters.get("style"))
 		        || "provider_radio".equals(parameters.get("style"));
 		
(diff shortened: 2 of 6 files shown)


omod/src/main/java/org/openmrs/module/htmlformentry/web/controller/ProviderSearchController.java
```diff
@@ -33,6 +33,10 @@ public Object getProviders(@RequestParam(value = "searchParam", required = false
 		
 		List<Provider> providerList = HtmlFormEntryUtil.getProviders(providerRoleIds, true);
 		
+		if (matchMode == null) {
```
chibongho, line 40: By default, the `<obs style="provider_autocomplete" />` element is rendered with `<input type='hidden' id='*_matchMode_hid' value='ANYWHERE' />` to indicate that the matchMode should be ANYWHERE. However, if the `<obs>` is conditionally rendered with `<controls><when></controls>`, the value of matchMode is reset to "". This gets around it by always defaulting matchMode to ANYWHERE if it's not specified.

mogoodrich: So if I'm understanding it correctly, you discovered that match mode doesn't work  when hidden within a controls/when?  

Now, if you specific another match mode rather than "ANYWHERE" but within a controls/when, does it revert to "ANYWHERE", or is it only a problem in the "default" case?

If you've confirmed there's no easy way to fix this, I'm okay with this workaround, but let's add some of the notes above as a comment in the code (and add a quick note to the text on the limitations--ie don't use match mode within a when/then).

mogoodrich (commented):
@chibongho generally looks good, but one question

omod/src/main/java/org/openmrs/module/htmlformentry/web/controller/ProviderSearchController.java
chibongho: Yes. If I put `<encounterProviderAndRole autocompleteProvider="true" providerMatchMode="START" />` inside a controls/when, the providerMatchMode value is erased. Without the providerMatchMode value,  the server does not do provider filtering at all, and just returns the entire list of providers regardless of what's typed in the text box. So, the providerMatchMode value erasure aside, I think we should add this default value anyway.

I haven't actually implemented `providerMatchMode` for `<obs style="provider_autocomplete">`, so it always defaults to ANYWHERE. 

I will:
- document the providerMatchMode defaulting to ANYWHERE for provider_autocomplete
- Add a comment explaining for why we need to default the matchMode value if it's not set
- file a ticket for fixing the `_matchMode_hid` value being erased when rendering inside a controls/when. (this might be a bigger issue for `<input type='hidden'>` in general)

chibongho: - documented the new `style="provider_autocomplete"` feature, and its behavior being equivalent to `providerMatchMode=ANYWHERE`, in [here](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25476979/HTML+Form+Entry+Module+Reference#Attributes.16).
- Updated with comment regarding the `<input>` value clearing bug
- Filed ticket for `<input>` value clearing bug [here](https://openmrs.atlassian.net/browse/HTML-887).

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
```diff
@@ -565,40 +567,55 @@ private void prepareWidgets(T context, Map<String, String> parameters) {
...
-					if (!providerOptions.isEmpty()) {
-						for (Option option : providerOptions) {
-							((SingleOptionWidget) valueWidget).addOption(option);
+					if ("provider_autocomplete".equals(parameters.get("style"))) {
+						MatchMode matchMode = MatchMode.ANYWHERE;
+						if (StringUtils.isNotBlank(parameters.get("providerMatchMode"))) {
+							try {
+								matchMode = MatchMode.valueOf(parameters.get("providerMatchMode").toUpperCase());
+							}
+							catch (IllegalArgumentException e) {
```
mseaton: Why are we catching an IllegalArgumentException here?  Shouldn't this catch "Exception?  And shouldn't it throw something like a BadFormDesignException (I forget the exact name)

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
```diff
@@ -282,6 +282,8 @@ public static Obs createObs(Concept concept, Object value, Date datetime, String
 			} else if (value instanceof Person) {
 				Person person = (Person) value;
 				obs.setValueText(person.getId().toString() + " - " + person.getPersonName().toString());
+			} else if (value instanceof Provider) {
+				obs.setValueText(((Provider) value).getId().toString());
```
mseaton, line 286: I don't love setting the primary key id to the value - it would be much better if this were the uuid - but if this is the pre-existing convention, I guess it is fine.  Can you confirm?

mseaton (approved):
Generally looks good- just really one comment regarding the Exceptions

api/src/main/java/org/openmrs/module/htmlformentry/element/ObsSubmissionElement.java
chibongho: Changed to catching Exception and throwing BadFormDesignException. I also had to add `throws BadFormDesignException` in various functions (or `throws Exception` for test functions), as it's a checked exception.

api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java
chibongho: So we have `"<id> - <name>"` format for Location and Person when saving them as obs values, but reading / parsing the value is more relaxed and can accept that, the id or the uuid.

However, `Provider` is stored as just `<id>`. We've been doing that for `<obs style="provider_dropdown">` and `<obs style="provider_radio">` already.

None of them are stored as uuid.

mogoodrich: Yeah, sounds like legacy behavior--if it's straightforward to change all of them to use uuids (while being backwards compatible) it would likely make sense to fix, though I my gut would be to do it in a quick follow-on ticket and commit so we can track it separately in case there are any issues.

chibongho: Filed https://openmrs.atlassian.net/browse/HTML-888 for this.

PR #338, opened 2026-06-10
HTML-886: Add new tag to record appointment service as obs

chibongho:
See: https://openmrs.atlassian.net/browse/HTML-886

Adds a `<AppointmentServiceObs>` tag to record appointment service as obs.

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
@@ -78,7 +78,8 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 		if (!specialityFilter.isEmpty()) {
 			services = services.stream()
 			        .filter(s -> s.getSpeciality() != null
-			                && specialityFilter.contains(s.getSpeciality().getName().toLowerCase().trim()))
+			                && (specialityFilter.contains(s.getSpeciality().getName().toLowerCase().trim())
+			                        || specialityFilter.contains(s.getSpeciality().getUuid().toLowerCase().trim())))
 			        .collect(Collectors.toList());
 		}
 
@@ -87,7 +88,7 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 
 		serviceWidget.addOption(new Option("", "", false));
 		for (AppointmentServiceDefinition svc : services) {
-			serviceWidget.addOption(new Option(svc.getName(), svc.getId() + " - " + svc.getName(), false));
+			serviceWidget.addOption(new Option(svc.getName(), svc.getUuid(), false));
 		}
 		if (existingValue != null) {
 			serviceWidget.setInitialValue(existingValue);
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
@@ -51,10 +51,20 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 			tagControlId = parameters.get("controlId");
 		}
 
+		boolean insideObsGroup = context.getCurrentObsGroupConcepts() != null
+		        && context.getCurrentObsGroupConcepts().size() > 0;
 		if (StringUtils.isNotBlank(tagControlId)) {
-			existingObs = context.getObsFromExistingObs(concept, tagControlId);
+			if (insideObsGroup) {
+				existingObs = context.getObsFromCurrentGroup(tagControlId);
+			} else {
+				existingObs = context.getObsFromExistingObs(concept, tagControlId);
+			}
 		} else {
-			existingObs = context.removeExistingObs(concept, (Concept) null);
+			if (insideObsGroup) {
+				existingObs = context.getObsFromCurrentGroup(concept, (Concept) null);
+			} else {
+				existingObs = context.removeExistingObs(concept, (Concept) null);
+			}
 		}
 		String existingValue = existingObs != null ? existingObs.getValueText() : null;
 
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
@@ -7,11 +7,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.StringUtils;
 import org.openmrs.Concept;
 import org.openmrs.Obs;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.appointments.model.AppointmentServiceDefinition;
 import org.openmrs.module.appointments.service.AppointmentServiceDefinitionService;
+import org.openmrs.module.htmlformentry.BadFormDesignException;
 import org.openmrs.module.htmlformentry.FormEntryContext;
 import org.openmrs.module.htmlformentry.FormEntryContext.Mode;
 import org.openmrs.module.htmlformentry.FormEntrySession;
@@ -29,22 +31,32 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 
 	private Obs existingObs;
 
-	private DropdownWidget serviceWidget;
+	DropdownWidget serviceWidget;
 
 	private ErrorWidget errorWidget;
 
-	public AppointmentServiceObsElement(FormEntryContext context, Map<String, String> parameters) {
+	private String tagControlId;
+
+	public AppointmentServiceObsElement(FormEntryContext context, Map<String, String> parameters) throws BadFormDesignException{
 		String conceptId = parameters.get("conceptId");
 		if (conceptId == null) {
-			throw new IllegalArgumentException("appointmentServiceObs requires a conceptId parameter");
+			throw new BadFormDesignException("appointmentServiceObs requires a conceptId parameter");
 		}
 		concept = HtmlFormEntryUtil.getConcept(conceptId);
 		if (concept == null) {
-			throw new IllegalArgumentException("Could not find concept: " + conceptId);
+			throw new BadFormDesignException("Could not find concept: " + conceptId);
 		}
 
-		existingObs = context.removeExistingObs(concept, (Concept) null);
-		String existingUuid = existingObs != null ? existingObs.getValueText() : null;
+		if (parameters.containsKey("controlId")) {
+			tagControlId = parameters.get("controlId");
+		}
+
+		if (StringUtils.isNotBlank(tagControlId)) {
+			existingObs = context.getObsFromExistingObs(concept, tagControlId);
+		} else {
+			existingObs = context.removeExistingObs(concept, (Concept) null);
+		}
+		String existingValue = existingObs != null ? existingObs.getValueText() : null;
 
 		List<String> specialityFilter = TagUtil.parseListParameter(parameters, "specialities", String.class).stream()
 		        .map(String::toLowerCase).map(String::trim).collect(Collectors.toList());
@@ -65,10 +77,10 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 
 		serviceWidget.addOption(new Option("", "", false));
 		for (AppointmentServiceDefinition svc : services) {
-			serviceWidget.addOption(new Option(svc.getName(), svc.getUuid(), false));
+			serviceWidget.addOption(new Option(svc.getName(), svc.getId() + " - " + svc.getName(), false));
 		}
-		if (existingUuid != null) {
-			serviceWidget.setInitialValue(existingUuid);
+		if (existingValue != null) {
+			serviceWidget.setInitialValue(existingValue);
 		}
 
 		context.registerWidget(serviceWidget);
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,110 @@
...
+		String conceptId = parameters.get("conceptId");
+		if (conceptId == null) {
+			throw new IllegalArgumentException("appointmentServiceObs requires a conceptId parameter");
+		}
+		concept = HtmlFormEntryUtil.getConcept(conceptId);
+		if (concept == null) {
+			throw new IllegalArgumentException("Could not find concept: " + conceptId);
+		}
+
+		existingObs = context.removeExistingObs(concept, (Concept) null);
```
mogoodrich: We should test/document what happens if you *multiple* of these tags on a single form with the same concept id.  At minimum, I don't think it will be able to reproduce them in the same order on reload.   We should likely add support for mapping via form namespace and path via the control Id as the obs tag does, see: https://openmrs.atlassian.net/browse/HTML-788  (or if we don't make a new tag but use the obs tag we should get this out of the box).

mogoodrich (changes requested):
Can you add some tests, to demo how this works?

What was the considering with adding this as a new tag, versus a "obs" with 
"style='appointmentService'" (like we have for location with "style='location')

(If this had all been discussed and decided while I was on vacay, that's fine, but if it hasn't been discussed, worth a discussion)

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: I don't think this will handle the rendering order correctly if we have multiple of these tags. That said, do we need to worry about that? An appointment can only be associated with one appointment service, and adding multiple of these tags for the same appointment should be a form design error.

chibongho:
> Can you add some tests, to demo how this works?
> 
> What was the considering with adding this as a new tag, versus a "obs" with "style='appointmentService'" (like we have for location with "style='location')
> 
> (If this had all been discussed and decided while I was on vacay, that's fine, but if it hasn't been discussed, worth a discussion)

Yeah, I can add some tests.

Speaking with @mseaton, I think the biggest motivation of having a separate tag is that `<obs>` is a bit of a kitchen sink with too many things in it.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
mogoodrich: But you could be scheduling multiple appointments (for different services) on this form, correct? Would that work?  This also ties into my previous question about why we decided to make this a new tag (instead of a obs tag with style='appointment')

mogoodrich:
> Speaking with @mseaton, I think the biggest motivation of having a separate tag is that `<obs>` is a bit of a kitchen sink with too many things in it.

Ah, okay, that makes sense then--we can keep as a separate tag.  I do think we may want to add support for "controlId" attribute, but why don't you start by adding some tests so I can understand it better and then we can discuss.  @chibongho

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,110 @@
...
+	private Obs existingObs;
+
+	private DropdownWidget serviceWidget;
+
+	private ErrorWidget errorWidget;
+
+	public AppointmentServiceObsElement(FormEntryContext context, Map<String, String> parameters) {
+		String conceptId = parameters.get("conceptId");
+		if (conceptId == null) {
+			throw new IllegalArgumentException("appointmentServiceObs requires a conceptId parameter");
```
mseaton: All of these `IllegalArgumentException` should probably be `BadFormDesignException`

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
mseaton: If we are going to put this tag in HFE, we should aim to support as many of the best-practice nuances that there are already established.  I'm fine putting this in the ObsSubmissionElement, but I'm 100% convinced this functionality belongs in the HFE module at all (versus in pihcore, etc)

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,110 @@
...
+		}
+		concept = HtmlFormEntryUtil.getConcept(conceptId);
+		if (concept == null) {
+			throw new IllegalArgumentException("Could not find concept: " + conceptId);
+		}
+
+		existingObs = context.removeExistingObs(concept, (Concept) null);
+		String existingUuid = existingObs != null ? existingObs.getValueText() : null;
```
mseaton: So in your other active PR, you used provider id, here you are using uuid.  I agree uuid is best, but why the inconsistency?

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,110 @@
...
+		context.registerWidget(serviceWidget);
+		context.registerErrorWidget(serviceWidget, errorWidget);
+	}
+
+	@Override
+	public String generateHtml(FormEntryContext context) {
+		StringBuilder sb = new StringBuilder();
+		sb.append(serviceWidget.generateHtml(context));
+		if (context.getMode() != Mode.VIEW) {
+			sb.append("<span class='required'>*</span>");
```
mseaton: Can you explain what this does?  Will this always mark this type of question as required?

mseaton (commented):
Generally looks good but see comments

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: It really ought to be required, but I haven't gotten a way to make it be conditionally required only if the user choose to create an appointment. What I have now is making this input technically optional in the backend, but have the subform [here](https://github.com/PIH/openmrs-config-pihsl/pull/322#issuecomment-4719609344) enforce the conditional requirement check via JS. Adding the `<span>` here is technically incorrect; I'll remove.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: I'll change it to the `"<id> - <serviceName>"` format similar to Location and Person.

mogoodrich: I'd be okay with keeping this as-is as this is a new domain we are capturing... I was more concerned about, say, capturing providers using "uuid" in some cases and "id" in others.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,138 @@
...
+		List<AppointmentServiceDefinition> services = Context.getService(AppointmentServiceDefinitionService.class)
+		        .getAllAppointmentServices(false);
+		services.sort((a, b) -> a.getName().compareToIgnoreCase(b.getName()));
+
+		if (!specialityFilter.isEmpty()) {
+			services = services.stream()
+			        .filter(s -> s.getSpeciality() != null
+			                && specialityFilter.contains(s.getSpeciality().getName().toLowerCase().trim()))
+			        .collect(Collectors.toList());
+		}
```
mogoodrich, line 84: Sorry I missed this in the initial review, but the speciality filter takes a list of speciality names (as opposed to uuids)?  If that's the general pattern used in Appointments, I'm okay with that, but we should be careful in our usage (and that any localization doesn't break things)

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,138 @@
...
+
+		serviceWidget = new DropdownWidget();
+		errorWidget = new ErrorWidget();
+
+		serviceWidget.addOption(new Option("", "", false));
+		for (AppointmentServiceDefinition svc : services) {
+			serviceWidget.addOption(new Option(svc.getName(), svc.getId() + " - " + svc.getName(), false));
```
mogoodrich: Sorry, the more I think about it, this should be uuid, regardless of what the other objects are doing... when reloading an existing obs is this doing a full match and so would fail on edit if the text name of a service as was changed?

mogoodrich (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: I followed Initializer, which accepts using the specifality name to identify speciality. I confirmed that the `appointment_speciality.name` in the DB has a unique constraint.

mseaton: I think it's fine to match on name, but matching on uuid would also be expected, I would think.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
mseaton: Yeah, this should definitely be uuid.  Is there no uuid on appointmentservicedefinition?

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -0,0 +1,138 @@
...
+			} else {
+				existingObs = context.getObsFromExistingObs(concept, tagControlId);
+			}
+		} else {
+			if (insideObsGroup) {
+				existingObs = context.getObsFromCurrentGroup(concept, (Concept) null);
+			} else {
+				existingObs = context.removeExistingObs(concept, (Concept) null);
+			}
+		}
```
mseaton, line 68: Presumably you took the above from the ObsSubmissionElement and replicated that here?  Not entirely sure what that is doing.

api/src/main/java/org/openmrs/module/htmlformentry/handler/AppointmentServiceObsTagHandler.java
```diff
@@ -0,0 +1,27 @@
...
+import org.openmrs.module.htmlformentry.BadFormDesignException;
+import org.openmrs.module.htmlformentry.FormEntrySession;
+import org.openmrs.module.htmlformentry.FormSubmissionController;
+import org.openmrs.module.htmlformentry.element.AppointmentServiceObsElement;
+
+public class AppointmentServiceObsTagHandler extends SubstitutionTagHandler {
+
+	@Override
+	protected List<AttributeDescriptor> createAttributeDescriptors() {
+		return Collections.singletonList(new AttributeDescriptor("conceptId", Concept.class));
```
mseaton, line 17: Do we also need to add in specialties here?

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
mogoodrich: Yeah, I didn't double-check, but this looks like the logic I wrote (or someone else?) that takes control id into account, if available, when finding a matching obs.  Assuming it follows the same pattern as ObsSubmissionElement, should (hopefully!) work out of the box.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
mogoodrich: Yes, I'm thinking it would be trivial to just extend this match to check both uuid and name, right?  (ie change line 81 to check if the specialityFilter contains either the name or uuid of the "s" value)

api/src/main/java/org/openmrs/module/htmlformentry/handler/AppointmentServiceObsTagHandler.java
chibongho: No. From Claude:
> `createAttributeDescriptors() `is used by the form exporter to resolve references to OpenMRS objects (concepts, locations, etc.) into UUIDs so they survive being imported on another server with different local IDs. The `specialities` attribute doesn't need this because it already accepts names or UUIDs directly in the XML.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: Yes, it's taken from `ObsSubmissionElement`. How we get the obs value is different depending on whether it's in an obsgroup.

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
chibongho: There is. I'll use uuid.

mseaton (approved):

PR #339, opened 2026-06-23
HTML-886 add attribute to specify sorting of options for appointmentS…

chibongho:
…erviceObs tag

This PR adds an optional `sortBy` attribute to the `<apointmentServiceObs>` tag. When specified with the value `specialitiesOrderThenServiceName`, it sorts the appointment service options first by the order of the specialities in the `specialities` attribute, then by service name. This allows us to make the "Other"  service option appear at the end of the dropdown list.

See: https://github.com/PIH/openmrs-config-pihsl/pull/324 for usage

### DIFF
--- a/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
@@ -1,10 +1,13 @@
 package org.openmrs.module.htmlformentry.element;
 
 import javax.servlet.http.HttpServletRequest;
+
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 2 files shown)


cioan (approved):
LGTM!

api/src/main/java/org/openmrs/module/htmlformentry/element/AppointmentServiceObsElement.java
```diff
@@ -37,6 +42,14 @@ public class AppointmentServiceObsElement implements HtmlGeneratorElement, FormS
 
 	private String tagControlId;
 
+	private static final String ATTRIBUTE_SPECIALITIES = "specialities";
+	private static final String ATTRIBUTE_SORT_BY = "sortBy";
+
+	private static final String ATTRIBUTE_SORT_BY_SPECIALITIES_ORDER_THEN_SERVICE_NAME = "specialitiesOrderThenServiceName";
```
mseaton, line 48: Not in love with this attribute name.  Can we change this so that sortBy can take a comma-delimited list of options, so the default would be "sortBy=serviceName", but in our forms it would be "sortBy=specialty,serviceName"?

PR #340, opened 2026-06-25
HTML-882 Create tag to schedule appointment

chibongho:
Resurrected https://github.com/openmrs/openmrs-module-htmlformentry/pull/333, which adds the `<ScheduleAppointment>` tag, with additional changes:
- make it store `appointment uuid` as a obs upon submission
- make it display the appointment in VIEW and EDIT mode.
- add `specialities` attribute to filter options in the appointment services dropdown (similar to the `<appointmentServiceObs>` tag)
- make options in the appointment services dropdown sortable with `sortServicesBy` attribute (similar to the `<appointmentServiceObs>` tag)
- add `controlId` attribute

https://github.com/user-attachments/assets/0483943b-c6f8-4cf2-a343-13fbdbb5153c


api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		        .stream().map(String::toLowerCase).map(String::trim).collect(Collectors.toList());
+		List<Speciality> allSpecialities = Context.getService(SpecialityService.class).getAllSpecialities();
+		List<String> specialityFilterUuids = specialityFilterNames.stream()
+		        .map(s -> allSpecialities.stream()
+		                .filter(sp -> sp.getName().equalsIgnoreCase(s) || sp.getUuid().equalsIgnoreCase(s))
+		                .findFirst())
+		        .filter(Optional::isPresent)
+		        .map(Optional::get)
+		        .map(Speciality::getUuid)
+		        .collect(Collectors.toList());
```
mogoodrich: We should just confirm this works if there are no specialities set up (assuming we want to support that)

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		// Build comparator from sortBy list (e.g. sortBy="specialityOrder,serviceName")
```
mogoodrich: Let's make sure we document in the wiki how this works, I'm not 100% following (but frankly partially that is because I'm not familiar with the appointment modelling and how services, types and specialities interact)

cioan (approved):
Looks good, @chibongho ! Thanks!

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		if (context.getMode() != Mode.ENTER) {
+			// VIEW/EDIT: claim the UUID obs so HFE doesn't warn about an unmatched obs,
+			// then fetch the appointment for display in generateHtml().
+			if (appointmentUuidConceptMapping != null) {
+				Concept uuidConcept = HtmlFormEntryUtil.getConcept(appointmentUuidConceptMapping);
+				if (uuidConcept != null) {
+					Obs uuidObs = StringUtils.isNotBlank(tagControlId)
+					        ? context.getObsFromExistingObs(uuidConcept, tagControlId)
+					        : (context.removeExistingObs(uuidConcept).stream().findFirst().orElse(null));
```
mseaton: Does this remove / claim the obs in both cases?   Does it need to?  Does this need to handle the case where multiple uuid obs might be found?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+	public ScheduleAppointmentElement(FormEntryContext context, Map<String, String> parameters) {
+		appointmentUuidConceptMapping = parameters.get("appointmentUuidConcept");
+		tagControlId = parameters.get("controlId");
+
+		if (context.getMode() != Mode.ENTER) {
+			// VIEW/EDIT: claim the UUID obs so HFE doesn't warn about an unmatched obs,
+			// then fetch the appointment for display in generateHtml().
+			if (appointmentUuidConceptMapping != null) {
+				Concept uuidConcept = HtmlFormEntryUtil.getConcept(appointmentUuidConceptMapping);
```
mseaton: Wouldn't it be an error case if the uuidConcept is not found or provided?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+			}
+			return;
+		}
+
+		// ENTER mode: set up all widgets
+		optional = "true".equalsIgnoreCase(parameters.get("optional"));
+		id = parameters.get("id");
+		if (optional) {
+			scheduleChoiceWidget = new TextFieldWidget();
+			context.registerWidget(scheduleChoiceWidget);
```
mseaton: Not entirely sure what value there is in only registering this widget conditionally?  Shouldn't it always get registered, but only used conditionally?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		// --- Location ---
+		locationWidget = new DropdownWidget();
+		locationErrorWidget = new ErrorWidget();
+		locationWidget.addOption(new Option(Context.getMessageSourceService().getMessage(
+		    "htmlformentry.scheduleAppointment.chooseLocation"), "", false));
+		String locationTagName = parameters.getOrDefault("locationTag", DEFAULT_LOCATION_TAG);
+		LocationTag locationTag = Context.getLocationService().getLocationTagByName(locationTagName);
+		List<Location> locations = locationTag != null
+		        ? Context.getLocationService().getLocationsByTag(locationTag)
+		        : Context.getLocationService().getAllLocations(false);
```
mseaton: Presumably these already come back sorted by name?  If so, we want to do that here.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		if ("true".equalsIgnoreCase(parameters.get("restrictToCurrentVisitLocation"))
+		        && context.getVisit() != null) {
+			locations = HtmlFormEntryUtil.removeLocationsNotEqualToOrDescendentOf(locations,
+			    ((Visit) context.getVisit()).getLocation());
+		}
+
+		for (Location loc : locations) {
+			locationWidget.addOption(new Option(loc.getName(), loc.getUuid(), false));
```
mseaton: @mogoodrich - do we have any special handling elsewhere for location names that need to be run through the message source, or do we normally just return the location name here?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		typeErrorWidget = new ErrorWidget();
+		if (allowedTypes.size() > 1) {
+			typeWidget.addOption(new Option(Context.getMessageSourceService().getMessage(
+			    "htmlformentry.scheduleAppointment.chooseType"), "", false));
+		}
+		for (AppointmentKind kind : allowedTypes) {
+			typeWidget.addOption(new Option(kind.getValue(), kind.name(), false));
+		}
+		if (allowedTypes.size() == 1) {
+			typeWidget.setInitialValue(allowedTypes.get(0).name());
```
mseaton: If there is only one, presumably we don't need a dropdown, but just use a hidden widget maybe?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		timeErrorWidget = new ErrorWidget();
+		context.registerErrorWidget(startDateTimeWidget, startDateTimeErrorWidget);
+		context.registerWidget(timeErrorWidget);
+
+		// --- All-day toggle (hidden field; value set by JS radio buttons) ---
+		allDayWidget = new TextFieldWidget();
+		context.registerWidget(allDayWidget);
+
+		// --- Duration ---
+		durationWidget = new TextFieldWidget();
```
mseaton: Is there anything to make this a numeric?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		// Note
+		sb.append(fieldRow("schedule-appointment-note",
+		    msg("htmlformentry.scheduleAppointment.note"),
+		    noteWidget.generateHtml(context)));
+
+		if (optional) {
+			sb.append("</div>");
+		}
+
+		sb.append("</div>");
```
mseaton, line 392: Didn't really review all this html concatenation - will need to come out in testing.  I would just ask that you look critically at the html output, make sure it looks good and things are aligned and spaced well, and that things are customizable via CSS as needed

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		return sb.toString();
+	}
+
+	private String generateViewEditHtml() {
+		if (existingAppointment == null) {
+			if (appointmentUuidConceptMapping == null) {
```
mseaton: What does this mean?  Isn't this an error state?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+			cal.setTime(start);
+			String timeStr = String.format("%02d:%02d",
+			    cal.get(Calendar.HOUR_OF_DAY), cal.get(Calendar.MINUTE));
+			sb.append(viewRow(msg("htmlformentry.scheduleAppointment.appointmentTime"), timeStr));
+		}
+
+		// Location
+		if (existingAppointment.getLocation() != null) {
+			sb.append(viewRow(msg("htmlformentry.scheduleAppointment.location"),
+			    existingAppointment.getLocation().getName()));
```
mseaton: Same question here about location name @mogoodrich

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+			    existingAppointment.getComments()));
+		}
+
+		// Status
+		if (existingAppointment.getStatus() != null) {
+			sb.append(viewRow(msg("htmlformentry.scheduleAppointment.status"),
+			    existingAppointment.getStatus().name()));
+		}
+
+		sb.append("</div>");
```
mseaton, line 460: Same comments here as I made for edit mode - please make this look as good  - concise and well spaced and aligned by default, and configurable via css/js by ensuring everything has appropriate classes and ids where needed.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		}
+
+		if (optional && !"yes".equals(scheduleChoiceWidget.getValue(context, submission))) {
+			return Collections.emptyList();
+		}
+
+		List<FormSubmissionError> errors = new ArrayList<>();
+
+		String locationUuid = (String) locationWidget.getValue(context, submission);
+		if (locationUuid == null || locationUuid.isEmpty()) {
```
mseaton: Use StringUtils.isBlank (and same for the below checks)

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		FormEntryContext context = session.getContext();
+
+		if (optional && !"yes".equals(scheduleChoiceWidget.getValue(context, submission))) {
+			return;
+		}
+
+		String locationUuid = (String) locationWidget.getValue(context, submission);
+		String serviceUuid = (String) serviceWidget.getValue(context, submission);
+		String typeValue = typeWidget.getOptions().size() == 1
```
mseaton: If this had been made a hidden element as I mentioned earlier if the size == 1, you could just get the value here in either case

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+		        ? typeWidget.getOptions().get(0).getValue()
+		        : (String) typeWidget.getValue(context, submission);
+		Date date = (Date) startDateTimeWidget.getDateWidget().getValue(context, submission);
+		String allDayValue = (String) allDayWidget.getValue(context, submission);
+		boolean isAllDay = "allday".equals(allDayValue);
+		Provider selectedProvider = (Provider) providerWidget.getValue(context, submission);
+		String note = (String) noteWidget.getValue(context, submission);
+
+		AppointmentServiceDefinition service = services.stream()
+		        .filter(s -> s.getUuid().equals(serviceUuid))
```
mseaton: Is there really no API method to get an appointment service definition by uuid?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+					// ignore invalid input
+				}
+			}
+			if (durationMins == null && service != null) {
+				durationMins = service.getDurationMins();
+			}
+			endDateTime = durationMins != null
+			        ? new Date(startDateTime.getTime() + (long) durationMins * 60 * 1000)
+			        : null;
+		}
```
mseaton, line 600: Didn't really review the above date/time stuff, will need testing

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,681 @@
...
+
+		Appointment appointment = new Appointment();
+		appointment.setPatient(session.getPatient());
+		appointment.setLocation(Context.getLocationService().getLocationByUuid(locationUuid));
+		appointment.setService(service);
+		if (typeValue == null || typeValue.isEmpty()) {
+			return;
```
mseaton: What is this all about???  This seems very wrong.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: It needs to be run through the message source, it looks like there is an HtmlFormEntryUtil.format() to do this. (See EncounterDetailsSubmissionElement)

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: See EncounterDetailsSubmissionElement to see how we sort there.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: Should be formatted, see above.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mogoodrich: The tagControlId should handle this--or rather the implementer should be expected to use controlId if they have multiple appointment tags on one form.

That being said, I didn't double-check that this matches the same pattern as in ObsSubmssionElement, we should confirm that--not sure if the stream stuff is correct. (I think see the "setExistingObs" method in ObsSubmissionElemet)

mogoodrich (changes requested):

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: Yeah... I'll make it check for that and throw `BadFormDesignException` regardless of `Mode`.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: `<ObsSubmssionElement>` calls `context.removeExistingObs(uuidConcept, (Concept) null)`, which does the same thing, but is more readable. I'll change it to that.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll add a sort with `OptionComparator`

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: The `   .filter(svc -> specialityFilterUuids.isEmpty() || ...` further down should make it so that if speciaiities is an empty list, we let the service show.

That said, I didn't check it against the case when the specialities is something invalid / mispelled. If we do `<scheduleAppointment specialities="invalid" />", it will show all services. I'll raise an exception when that happens.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: It allows sorting of services based on the order in which we list the specialities. So if we do `<SchedulAppointment specialities="maternal,other" orderServicesBy="specialityOrder" />`, then it lists the maternal services first, before the "other" services.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll swithc it to use NumberFieldWidget.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I'll move the check into the constructor

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: There is, I'll change it.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: I added classes to the HTML tags and also separated the inline styling into CSS styles.

chibongho:
I think this is ready for review again. I updated the tag's documentation here:

https://openmrs.atlassian.net/wiki/spaces/docs/pages/25476979/HTML+Form+Entry+Module+Reference#%3CscheduleAppointment%3E

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+			if (uuidObs != null && uuidObs.getValueText() != null) {
+				try {
+					existingAppointment = Context.getService(AppointmentsService.class)
+					        .getAppointmentByUuid(uuidObs.getValueText());
+				}
+				catch (Exception e) {
+					// appointments module unavailable
+				}
+			}
+			return;
```
mseaton: Not a huge fan of return statements in the middle of a method.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
mseaton: Did this get addressed @chibongho ?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+		sb.append("<style>");
+		sb.append(".schedule-appointment p{margin:0.3em 0;}");
+		sb.append(".schedule-appointment p>label{display:block;margin-bottom:0.2em;}");
+		sb.append(".schedule-appointment .field-widget{display:flex;align-items:center;gap:0.25em;flex-wrap:wrap;}");
+		sb.append(".schedule-appointment-choice label{display:inline-flex;align-items:center;gap:0.3em;cursor:pointer;margin-right:1em;}");
+		sb.append(".schedule-appointment-timing-options{display:flex;gap:1.5em;align-items:center;}");
+		sb.append(".schedule-appointment-timing-option{display:inline-flex;align-items:center;gap:0.3em;cursor:pointer;}");
+		sb.append(".schedule-appointment-date-time-widget{display:inline-flex;align-items:center;gap:0.5em;flex-wrap:wrap;}");
+		sb.append(".schedule-appt-time{display:inline-flex;align-items:center;gap:0.25em;}");
+		sb.append("</style>");
```
mseaton, line 295: This could be fine, but I worry a bit whether this will be easy or difficult to override if it is styled in the page vs in an included css file.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+				// non-numeric input — falls through to error
+			}
+			if (durNum == null || durNum.intValue() < 1) {
+				errors.add(new FormSubmissionError(context.getFieldName(durationErrorWidget),
+				        msg("htmlformentry.scheduleAppointment.error.durationRequired")));
+			}
+		}
+
+		Provider provider = (Provider) providerWidget.getValue(context, submission);
+		if (provider == null) {
```
mseaton, line 527: Is provider always a required field when creating an appointment?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+
+		// Build end datetime
+		Date endDateTime;
+		if (isAllDay) {
+			Calendar endCal = Calendar.getInstance();
+			endCal.setTime(date);
+			endCal.set(Calendar.HOUR_OF_DAY, 23);
+			endCal.set(Calendar.MINUTE, 59);
+			endCal.set(Calendar.SECOND, 59);
+			endCal.set(Calendar.MILLISECOND, 999);
```
mseaton: I don't think setting millisecond is generally what we do.  There is also a utility method you can use for this I believe - `OpenmrsUtil.lastSecondOfDate`

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+		Appointment appointment = new Appointment();
+		appointment.setPatient(session.getPatient());
+		appointment.setLocation(Context.getLocationService().getLocationByUuid(locationUuid));
+		appointment.setService(service);
+		appointment.setAppointmentKind(AppointmentKind.valueOf(typeValue));
+		appointment.setStartDateTime(startDateTime);
+		appointment.setEndDateTime(endDateTime);
+		appointment.setComments(note);
+
+		if (selectedProvider != null) {
```
mseaton: Per my comment above, in the validation Provider is required.  Probably need to loosen that there?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+			appointmentProvider.setAppointment(appointment);
+			Set<AppointmentProvider> providers = new HashSet<>();
+			providers.add(appointmentProvider);
+			appointment.setProviders(providers);
+		}
+
+		session.getSubmissionActions().addAppointmentToCreate(appointment);
+
+		// Save the appointment UUID as a single obs so VIEW/EDIT mode can retrieve it
+		if (appointmentUuidConceptMapping != null) {
```
mseaton: Don't you already do all of this in the Constructor?  By the time you get here, you should be able to already have validated that this exists and should be able to keep it around as a variable, right?

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+		}
+		if (end == null) {
+			return true;
+		}
+		cal.setTime(end);
+		return cal.get(Calendar.HOUR_OF_DAY) == 23 && cal.get(Calendar.MINUTE) == 59;
+	}
+
+	private String msg(String key) {
+		return Context.getMessageSourceService().getMessage(key);
```
mseaton: FYI - there is already an `HtmlFormEntryUtil.translate(String)` method.

api/src/main/resources/messages.properties
```diff
@@ -189,3 +189,31 @@ htmlformentry.conditionui.inactive.label=Inactive
...
+htmlformentry.scheduleAppointment.doSchedule=Schedule appointment
+htmlformentry.scheduleAppointment.skip=Skip for now
+htmlformentry.scheduleAppointment.viewModeMessage=To view or manage appointments for this patient, use the Appointments app.
+htmlformentry.scheduleAppointment.error.locationRequired=Location is required
+htmlformentry.scheduleAppointment.error.serviceRequired=Service is required
+htmlformentry.scheduleAppointment.error.typeRequired=Appointment type is required
+htmlformentry.scheduleAppointment.error.dateRequired=Appointment date is required
+htmlformentry.scheduleAppointment.error.timeRequired=Appointment time is required
+htmlformentry.scheduleAppointment.error.durationRequired=Duration is required
+htmlformentry.scheduleAppointment.error.providerRequired=Provider is required
```
mseaton, line 219: I would expect translations for this in French and Spanish at minimum, possibly `ht` as well.

mseaton (approved):
Approving, as I think it generally looks good, though there are a few comments I have made and changes I think you should apply before you merge, but once you do so I'm happy for this to move forward and into the testing phase.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+			comparator = comparator == null ? next : comparator.thenComparing(next);
+		}
+
+		services = Context.getService(AppointmentServiceDefinitionService.class).getAllAppointmentServices(false);
+		final Comparator<AppointmentServiceDefinition> finalComparator = comparator;
+		services = services.stream()
+		        .filter(svc -> specialityFilterUuids.isEmpty()
+		                || (svc.getSpeciality() != null && specialityFilterUuids.contains(svc.getSpeciality().getUuid())))
+		        .sorted(finalComparator)
+		        .collect(Collectors.toList());
```
dkayiwa: Nothing in the test suite exercises the speciality features this PR adds. The test dataset has one speciality and one service, and every test either omits `specialities`/`sortServicesBy` or takes the default path, so the speciality filter, the `specialityOrder` comparator, and the unknown-speciality `BadFormDesignException` on line 189 only ever run in production. That's the most intricate logic in the tag and it's the kind that regresses quietly, so it's worth adding regression coverage: a `specialities` value that narrows the service list, `sortServicesBy="specialityOrder"` ordering, and the unknown-speciality error path. Non-blocking.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -0,0 +1,685 @@
...
+
+	private ErrorWidget providerErrorWidget;
+
+	// Config params
+	private String appointmentUuidConceptMapping;
+
+	private String tagControlId;
+
+	// stored for duration lookup at submission time
+	private List<AppointmentServiceDefinition> services = new ArrayList<>();
```
dkayiwa: This field is only read inside the constructor (to build the service dropdown). `handleSubmission` re-fetches the service through `getAppointmentServiceByUuid` and uses that for the duration lookup, so the "stored for duration lookup at submission time" comment no longer holds and the field can just be a local variable in the constructor. Tiny cleanup, not important.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: Yeah, I added a hidden input when the size is 1. I can remove this `if`.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: The appointments module technically allows for NULL provider, but I'm following the O3 appointment app's behavior to require it.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: yeah, I can get rid of the if check here.

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
chibongho: yes, I can get rid of the if check here too

chibongho:
> Did this get addressed @chibongho ?

yeah, I made it a hidden input if there is only one type. I can get rid of that if check.

api/src/main/resources/messages.properties
mogoodrich: @chibongho @mseaton I believe we are still doing translations via Transifex for HTML Form Entry (do you remember, @dkayiwa?).

For that pattern, you would just add the English translations here, and then perform all the localization in Transifex--but that would be a big pain for this commit.  Let's see what happens when we merge, it's possible that Transifex will blow away all the translations you  added and we will have to force push them up.

(So no changes required now, but we will need to monitor after the merge)

mogoodrich (approved):

PR #341, opened 2026-07-16
HTML-889 make date widget field value clearable

chibongho:
Currently, the Date widget renders a readonly text input field that opens a DatePicker on click. This means that the widget does not allow for entering empty value on the Date once the user has clicked on that field. This PR makes the text input field not readonly, so that it's possible for the user to clear the field, and also add visual indication if the input date is invalid.

https://github.com/user-attachments/assets/ac380634-2d36-45c7-8099-d70c4c971330


mseaton (commented):
What I was expecting here was a little "x" icon to the right on the date picker box which would clear the inputs.  Not that this isn't a valid solution, but I worry a little bit about making this non-readonly.

cioan:
@chibongho , I agree with @mseaton . We should have an X next to the date picker. Maybe similar to the date picker we use for editing visit dates?

Screenshot 2026-07-16 at 2 51 23 PM

mogoodrich (commented):
Agree that having an "x" would be preferrable.

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -649,9 +649,28 @@ function setupDatePicker(jsDateFormat,yearsrange, jsLocale, displaySelector, val
 		setDatePickerValue(displaySelector, initialDateYMD);
 
 	// register a handler to set the date value to zero if the display widget is emptied (workaround for jquery bug http://bugs.jqueryui.com/ticket/5734)
+	// also highlight the display field with the same illegalValue class used by numeric fields if what's typed isn't
+	// a valid date, and forward the raw text into the hidden field so the server's strict date parser rejects it on
```
dkayiwa, line 653: The strict-parser backstop described here holds for obs and encounterDate (HtmlFormEntryUtil.convertToType parses yyyy-MM-dd with setLenient(false)), but not for the order widget, which reuses setupDatePicker via orderWidget.js enableDateWidgets. Order dates go through TagUtil.parseValue, whose SimpleDateFormat is left lenient, so an ISO-shaped but impossible date forwarded by this handler rolls over instead of failing: paste 2026-02-31 into a scheduled-date field and the order is saved for 2026-03-03 with no error, the red highlight being the only hint (and this comment explicitly frames the highlight as advisory). Since the design depends on the server rejecting what the client merely flags, I'd add setLenient(false) in TagUtil.parseValue as part of this change. That also tightens tag-attribute parsing, which goes through the same method and seems equally right to have strict.

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -649,9 +649,28 @@ function setupDatePicker(jsDateFormat,yearsrange, jsLocale, displaySelector, val
...
+	// submit the same way an unparseable number is rejected today - the highlight is just an early hint, not a hard block
 	jq.change(function () {
```
dkayiwa, line 655: The hidden field's change event now fires only for calendar picks (onSelect above triggers it explicitly); a typed, pasted, or cleared date updates the hidden field with no event at all, since the keyup sync and setDatePickerValue both write it with plain .val(). The in-repo consumer of that event is encounterDate's disallowMultipleEncountersOnDate, wired as an onChange attribute on the hidden input (EncounterDetailSubmissionElement.java:108) with no server-side backstop. So on a form using disallowMultipleEncountersOnDate="block", typing the date instead of picking it skips the duplicate-encounter check entirely and the duplicate encounter is created without any warning. Before this PR every UI path went through onSelect, so the check always ran. I'd centralize the notification in this handler: remember the last hidden value this handler announced (jq.data works for that), trigger the hidden field's change when it differs, and drop the explicit valueSelector trigger from onSelect so picks don't fire it twice. Comparing against the hidden field's current value isn't enough, because the keyup sync has usually already written it by the time change fires here.

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -649,9 +649,28 @@ function setupDatePicker(jsDateFormat,yearsrange, jsLocale, displaySelector, val
...
-		if (jq.val() == null || jq.val() == '') {
+		const val = jq.val();
+		if (val == null || val == '') {
+			jq.removeClass('illegalValue');
 			setDatePickerValue(displaySelector, null);
+		} else {
+			let parsed = null;
+			try {
+				parsed = jQuery.datepicker.parseDate(jsDateFormat, val);
```
dkayiwa, line 663: This parses with jQuery's global month and day names (English), but the picker instance may be running with the localized ones applied at line 647. jsDateFormat() explicitly supports month-name formats (MMMM/MMM become MM/M), so on a site with locale fr and a dd-MMM-yyyy date format, picking February 16 from the calendar puts "16-Fév-2026" in the display, this call throws on "Fév", the field turns red, and the else branch below overwrites the correct "2026-02-16" the datepicker had just written to the hidden field with the raw display text. The server then rejects that on submit, so a calendar-picked date can't be saved at all. It only bites in months whose localized name differs from English (Fév yes, Jul no), which will make it a miserable one to reproduce from a bug report. Same story for es, it and pt, and for correctly-typed dates in those locales. Passing the instance's own settings closes it:

```suggestion
				parsed = jQuery.datepicker.parseDate(jsDateFormat, val, {
					monthNames: jq.datepicker('option', 'monthNames'),
					monthNamesShort: jq.datepicker('option', 'monthNamesShort'),
					dayNames: jq.datepicker('option', 'dayNames'),
					dayNamesShort: jq.datepicker('option', 'dayNamesShort')
				});
```

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -649,9 +649,28 @@ function setupDatePicker(jsDateFormat,yearsrange, jsLocale, displaySelector, val
...
+			} catch (err) {
+				// not a valid date in the expected format
+			}
+			
+			if (parsed) {
+				jq.removeClass('illegalValue');
+			} else {
+				jq.addClass('illegalValue');
+				jQuery(valueSelector).val(val);
+			}
```
dkayiwa, line 673: The valid branch needs to write the hidden field too, otherwise the display and the submitted value can disagree. Right now it only clears the highlight and relies on the datepicker's keyup handler to have synced the hidden field, but keyup never fires for a context-menu paste or a drag-and-drop. In those cases this handler sees a parseable value and removes the highlight, and the form submits whatever the hidden field held before: on a new form the date is silently dropped, and in edit mode the old date is silently kept while the screen shows the corrected one. Nothing throws, so the first person to notice is whoever reads the record. This is the same failure HTML-852 reported before the field was made readonly ("upon submit, HTML Form Entry ignores that value... without throwing any kind of warning"), so I'd close it before merging:

```suggestion
			if (parsed) {
				jq.removeClass('illegalValue');
				jQuery(valueSelector).val(jQuery.datepicker.formatDate('yy-mm-dd', parsed));
			} else {
				jq.addClass('illegalValue');
				jQuery(valueSelector).val(val);
			}
```

The 'yy-mm-dd' matches the altFormat configured above, and formatting the parsed date rather than echoing the raw text also normalizes things like 2-digit years.

chibongho:
yeah, I'll just redo this with an "x" button and keep the input readonly. Less input validation to worry about.

PR #342, opened 2026-07-20
HTML-889 make date widget field value clearable

chibongho:
Adds an "x" button next to the date widget for clearing the date field.


api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java
```diff
@@ -111,7 +111,11 @@ public String generateHtml(FormEntryContext context) {
 			}
 			
 			if (!hidden) {
-				sb.append("<input type=\"text\" readonly size=\"10\" id=\"").append(fieldName).append("-display\"/>");
+				sb.append("<input type=\"text\" readonly size=\"10\" id=\"").append(fieldName)
+					.append("-display\" style=\"display:inline-block;vertical-align:middle;\"/>");
+				sb.append("<span class=\"ui-icon ui-icon-close\" style=\"display:inline-block;vertical-align:middle;cursor:pointer;\" ")
+					.append("onclick=\"clearDatePickerValue('#").append(fieldName).append("-display', '#")
+					.append(fieldName).append("')\" title=\"clear\"></span>");
```
dkayiwa: The build is red on both Java 8 and 11, and this change is the cause. `ZonedDateTimeWidget` extends `DateWidget` and renders the date part through `super.generateHtml()`, so the exact-string assertion in `ZonedDateTimeWidgetTest.generateHtml_shouldDisplaySubmittedDateAsUTCInEditMode` now fails: the new `style="..."` on the display input and the `<span class="ui-icon ui-icon-close">` show up in the output it compares against, and the reactor stops at htmlformentry-api until that expected string is updated.

Separately, and optional: `DateWidgetTest` only covers the hidden path (`setHidden(true)`), so nothing directly asserts the visible edit-mode output where this button renders. A non-hidden case checking that the clear span is emitted would give the new behavior real coverage.

omod/src/main/webapp/resources/htmlFormEntry.js
```diff
@@ -669,6 +669,12 @@ function setDatePickerValue(displaySelector, ymd) {
 	}
 }
 
+function clearDatePickerValue(displaySelector, valueSelector) {
+	setDatePickerValue(displaySelector, null);
+	jQuery(valueSelector).change();
+	jQuery(displaySelector).change();
+}
```
dkayiwa, line 676: Worth confirming the scope here. `DateWidget.generateHtml()` is also the date half of `DateTimeWidget` and `ZonedDateTimeWidget` (composition in one, `super` in the other), so this button now shows up on datetime fields too. It's wired only to the date selectors, so clicking it clears the date and leaves the time dropdowns (and, on the zoned widget, the timezone) still set. Is clearing only the date what you intended there, or should the whole value reset? No crash risk either way, since an empty date was already reachable by never picking one, I just want to make sure the datetime case was considered.

api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java
dkayiwa: `title="clear"` is hardcoded English, while the other widgets localize user-facing text through `Context.getMessageSourceService().getMessage(...)` (`AutocompleteWidget`, `LocationWidget`, and others), so a message code would be more consistent and translatable. And because the control is a `<span>` with an `onclick`, keyboard and screen-reader users can't activate it, so a `<button type="button">` (or at least a `tabindex` and `role`) would make it reachable. Both are minor, fine to skip.

mogoodrich (commented):
Generally looks good, but see the comments from @dkayiwa

api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java
chibongho: Fixed the build. Also added a test in `DateWidgetTest` to cover non-hidden path.

omod/src/main/webapp/resources/htmlFormEntry.js
chibongho: Good catch. I didn't account for the DateTimeWidget. The UX was actuallly not great, as it's possible to provide a time without providing a date. I changed it so that the time (hour and minute) fields are disabled until a date has bee inputted in the Date widget. 

https://github.com/user-attachments/assets/859bcc92-807e-49f4-a5d8-a3f0efe68ca2

api/src/main/java/org/openmrs/module/htmlformentry/widget/DateWidget.java
chibongho: Changed "clear" to be localized. 

I tried to make turn the span into a button if it didn't render correctly. I'll leave it as is for now.

api/src/main/java/org/openmrs/module/htmlformentry/widget/DateTimeWidget.java
```diff
@@ -43,9 +44,16 @@ public String generateHtml(FormEntryContext context) {
...
 		String dateHTML = dateWidget.generateHtml(context);
 		String timeHTML = timeWidget.generateHtml(context);
-		
-		// combine them and return them
-		return dateHTML + " " + timeHTML;
+
+		StringBuilder sb = new StringBuilder();
+		sb.append(dateHTML).append(" ").append(timeHTML);
+
+		if (context.getMode() != Mode.VIEW && !dateWidget.isHidden()) {
```
dkayiwa: Worth also checking `timeWidget.isHidden()` here, though no in-repo tag hits the gap today. A hidden TimeWidget renders value-carrying hidden inputs (that is how a form fixes the time while collecting only the date), and the public two-arg constructor lets a module pair one with a visible date widget. The name selectors in `syncTimeFieldsToDate` match those hidden inputs too, so clearing the date zeroes and disables them, and once a date is re-picked they submit as 0, quietly turning the fixed time into midnight. The pairings in this repo are safe (ObsSubmissionElement and ScheduleAppointmentElement never hide the time, and ZonedDateTimeWidget hides both halves together), so take this or leave it:

```suggestion
		if (context.getMode() != Mode.VIEW && !dateWidget.isHidden() && !timeWidget.isHidden()) {
```

dkayiwa (approved):

PR #343, opened 2026-07-22
HTML-891: HTML Form Entry API module should not depend on OpenMRS Cor…

mogoodrich:
…e Web module

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -22,6 +22,10 @@
 			<artifactId>metadatamapping-api</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.openmrs.api</groupId>
 			<artifactId>openmrs-api</artifactId>
 			<classifier>tests</classifier>
@@ -32,10 +36,6 @@
 			<artifactId>openmrs-test</artifactId>
 			<type>pom</type>
 			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.openmrs.web</groupId>
-			<artifactId>openmrs-web</artifactId>
 		</dependency>
 
 
--- a/api/src/main/java/org/openmrs/module/htmlformentry/widget/AutocompleteWidget.java
+++ b/api/src/main/java/org/openmrs/module/htmlformentry/widget/AutocompleteWidget.java
@@ -3,9 +3,9 @@ package org.openmrs.module.htmlformentry.widget;
 import java.util.Iterator;
 import java.util.List;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.htmlformentry.FormEntryContext;
-import org.springframework.web.util.HtmlUtils;
 
 /**
  * A single option auto complete widget which provides auto complete suggestions using a list of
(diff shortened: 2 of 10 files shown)


pom.xml
```diff
@@ -116,6 +118,15 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <!-- spring-web is a plain library dependency (not the OpenMRS web-tier module) used directly
+                 by api-tier code, e.g. for MultipartHttpServletRequest and UriComponentsBuilder. Version
```
dkayiwa: `UriComponentsBuilder` isn't used anywhere in the module (I grepped the whole tree), so this example is a little misleading. The spring-web classes the api tier actually pulls in are `MultipartFile` / `MultipartHttpServletRequest` (in HtmlFormEntryUtil) and `HtmlUtils` / `JavaScriptUtils` (FormEntrySession, AutocompleteWidget, LocationWidget). Might be worth pointing the comment at one of those instead. Very minor though, feel free to ignore.

api/pom.xml
```diff
@@ -21,6 +21,10 @@
 			<groupId>org.openmrs.module</groupId>
 			<artifactId>metadatamapping-api</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+		</dependency>
```
mseaton, line 27: Do we want to include spring-web here?  Not sure if this is problematic?  What is this actually providing, anything?

pom.xml
```diff
@@ -16,6 +16,8 @@
         <openMRSVersion>2.8.0</openMRSVersion>
         <openMRSMinorVersion>2.8</openMRSMinorVersion>
         <legacyuiVersion>2.0.0</legacyuiVersion>
+        <!-- kept in sync with the Spring version openmrs-api/${openMRSVersion} depends on -->
+        <springVersion>5.3.30</springVersion>
```
mseaton, line 20: Why are we adding this?

pom.xml
mseaton: Are the usages of these thing able to be moved to the omod layer instead of adding this in here?

mseaton (approved):
It's fine.  I don't love it, but will work until we can do a proper reevaluation someday of what classes actually belong in the api layer in this module (eg. most likely does not)

api/src/main/java/org/openmrs/module/htmlformentry/widget/AutocompleteWidget.java
```diff
@@ -3,9 +3,9 @@
 import java.util.Iterator;
 import java.util.List;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.htmlformentry.FormEntryContext;
-import org.springframework.web.util.HtmlUtils;
```
mogoodrich, line 8: So there were scattered cases of using util methods like this from the springframework that Claude was able to identify and replace with apache.commons libraries

pom.xml
mogoodrich: I've add Claude update the comment below with the one remaining thing that it thinks will be difficult to untangle.

mogoodrich (commented):
So updated @dkayiwa  @mseaton  to remove a few more spring-web stuff in the API module that Claude found easy to swap out... I left (with a comment) the one think it thinks would be harder to remove (MultipartHttpServletRequest/MultipartFile)... didn't want to go too much farther down the rabbit hole but wanted to get it correctly documented.  Will merge this all in if the buld passes.

PR #344, opened 2026-07-22
HTML-882: drop support for all-day appointments in ScheduleAppointment tag

chibongho:
"All-day Appointment" is not well-supported by the appointments module, despite that being an option in the O3 appointments app (via config). This PR removes the ability to schedule all-day appointments in the `<scheduleAppointments>` element, since we don't need it anyway.


api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -332,28 +322,7 @@ public String generateHtml(FormEntryContext context) {
 			        .append("\" value=\"").append(typeWidget.getOptions().get(0).getValue()).append("\"/>");
 		}
 
-		// All-day / Specific-time toggle — shown BEFORE the date/time row
-		String allDayFieldName = context.getFieldName(allDayWidget);
-		sb.append("<p class=\"schedule-appointment-timing\">");
-		sb.append("<label>").append(msg("htmlformentry.scheduleAppointment.timing"))
```
dkayiwa, line 338: With the toggle gone, the `htmlformentry.scheduleAppointment.timing` and `htmlformentry.scheduleAppointment.specificTime` keys aren't referenced from anywhere anymore, but they're still in messages.properties and the es/fr/ht bundles. Worth dropping them in this PR so the dead strings don't stick around in the translation files. (`allDay` is still used by the view-mode rendering, so that one stays.)

api/src/main/java/org/openmrs/module/htmlformentry/element/ScheduleAppointmentElement.java
```diff
@@ -500,27 +467,22 @@ public Collection<FormSubmissionError> validateSubmission(FormEntryContext conte
...
-		String allDayValue = (String) allDayWidget.getValue(context, submission);
-		boolean isAllDay = "allday".equals(allDayValue);
-		if (!isAllDay) {
-			Date time = (Date) startDateTimeWidget.getTimeWidget().getValue(context, submission);
-			if (time == null) {
-				errors.add(new FormSubmissionError(context.getFieldName(timeErrorWidget),
-				        msg("htmlformentry.scheduleAppointment.error.timeRequired")));
-			}
+		Date time = (Date) startDateTimeWidget.getTimeWidget().getValue(context, submission);
+		if (time == null) {
```
dkayiwa, line 474: Could we add a validation test that omits the time selects (and one omitting duration), along the lines of `shouldRequireServiceInlineNotAsGlobalError`? Making time and duration unconditionally required is the one behavior this PR changes, and every submission test in the suite supplies both, so nothing would fail if the requirement quietly regressed. Not a blocker.

chibongho:
Added a couple tests on the time and duration fields, and removed the unused translation strings.

dkayiwa (approved):

PR #349, opened 2026-08-28
HTML-893: Obs checkbox element should only show dateLabel when checked

chibongho:
Before:
image

After:
image

image


mseaton (approved):
Looks ok to me, will need some testing in view, edit, and enter modes.

